PR #40, opened 2015-03-24
New spec

zachmullen:
This topic will make romanesco support the new and improved task specification.


romanesco/tasks/workflow.py
```diff
@@ -0,0 +1,123 @@
+import romanesco
+
+from romanesco.utils import toposort
+
+
+def run(task, inputs, outputs, task_inputs, task_outputs):
```
jeffbaumes: What is the intent of "inputs" and "outputs" as opposed to "task_inputs" and "task_outputs"? Not sure why both are needed.

zachmullen: I just passed all the vars that were referenced by the internals of each mode, we can refactor the implementations so we don't have to pass all of these to the functions.

jeffbaumes: Ok got it.

zachmullen:
@jeffbaumes This includes all of the spec refactoring necessary to allow use of romanesco as the COVALIC worker. The main remaining blocker there is the creation of the docker execution mode, which should be on a different topic.

One area I didn't make conform to the new spec yet is the workflow connection spec changes you proposed (which mostly seems to be changing field names). Do you want me to make those changes on this branch or should we cut it off here?

zachmullen:
With a few minor tweaks, it seems this now works with the existing tangelohub code base.

There's a lot more still to do but I think we should defer to future topics.

romanesco/__init__.py
```diff
@@ -199,258 +152,97 @@ def run(analysis, inputs, outputs=None, auto_convert=True, validate=True):
...
         the name of the output and ``binding`` is an output binding of the form
-        ``{"format": format, "data": data}``. If the `outputs` parameter
+        ``{"format": format, "data": data}``. If the `outputs` param
         is specified, the formats of these bindings will match those given in
         `outputs`. Additionally, ``"data"`` may be absent if an output URI
         was provided. Instead, those outputs will be saved to that URI and
         the output binding will contain the location in the ``"uri"`` field.
     """
+    def extractId(spec):
+        return spec["id"] if "id" in spec else spec["name"]
```
jeffbaumes: Ah, I see this is how you are maintaining backwards compatibility. Was wondering how it worked against unmodified tangelohub. Nice.

jeffbaumes:
Working for me against tangelohub master. LGTM!

zachmullen:
@jeffbaumes This works with tangelohub master for stuff like strings that are passed directly, but for files that get downloaded to the worker, the following PR is required to be compatible with these changes: https://github.com/Kitware/tangelohub/pull/100

PR #41, opened 2015-03-27
Clean up REST API

zachmullen:


jeffbaumes:
Before I test this - is this meant to function identically as the prior api? I will want to test this against the tangelohub app.

zachmullen:
Not identically, but the changes are subtle

* `romanescoCreateModule` was changed from `access.public` to `access.user`, since attempting to create a collection with `creatot=None` raises exceptions in core. If we want to allow collection creation by anonymous users, it would require some tweaks to core (though I'm not sure it's a good idea).
* I changed some places that were catching exceptions and returning objects that contained error messages. Instead I have them propagate the exception, which will cause Girder's REST framework to automatically generate an appropriate JSON error message and HTTP status code. If there was some reason we wanted to return a 200 in an error condition, we can revert that, although I think it's better this way, and makes the code cleaner.

jeffbaumes:
It makes sense that `romanescoCreateModule` has `user` access. I added this because I think it is currently an admin privilege to create a collection and I needed other users to be able to perform the operation.

Better error codes are good. TangeloHub may expect some errors to be returned in other ways but that should be trivial to fix.

Looks like I can try this out against TangeloHub and see if it works properly.

jeffbaumes:
With this branch and tangelohub on https://github.com/Kitware/tangelohub/pull/100 I get the following error when attempting to run a simple task through tangelohub. Task is

```json
{
    "inputs": [
        {
            "format": "number",
            "name": "a",
            "type": "number"
        },
        {
            "format": "number",
            "name": "b",
            "type": "number"
        }
    ],
    "mode": "python",
    "name": "Test",
    "outputs": [
        {
            "format": "number",
            "name": "c",
            "type": "number"
        }
    ],
    "script": "c = a + b"
}
```

Error in romanesco is
```
[2015-04-03 09:35:21,476: ERROR/MainProcess] Task romanesco.run[ad6c23dc-7943-470d-b3e5-b6ecc673248f] raised unexpected: KeyError('url',)
Traceback (most recent call last):
  File "/Library/Python/2.7/site-packages/celery/app/trace.py", line 240, in trace_task
    R = retval = fun(*args, **kwargs)
  File "/Library/Python/2.7/site-packages/celery/app/trace.py", line 437, in __protected_call__
    return self.run(*args, **kwargs)
  File "/code/romanesco/romanesco/__main__.py", line 13, in run
    url = kwargs.pop('url')
KeyError: 'url'
```

Launched the romanesco endpoint
```
http://localhost:9080/girder/api/v1/item/551e91d9358ebf13c3562772/romanesco
```

With a POST with payload containing
```json
{"inputs":{"a":{"type":"number","format":"number","data":1},"b":{"type":"number","format":"number","data":1}},"outputs":{"c":{"type":"number","format":"number"}}}
```

zachmullen:
https://github.com/Kitware/tangelohub/pull/100 actually supports changes on both this and #40 that was just merged. I'll merge master into this topic and it should work then.

jeffbaumes:
Ah I see. I'll retry once you merge in master.

zachmullen:
Merged master in

jeffbaumes:
Appears to be working correctly for me now. LGTM.

PR #47, opened 2015-04-06
Docker execution mode

zachmullen:


romanesco/tasks/docker.py
```diff
@@ -0,0 +1,101 @@
...
+
+    outputs['_stdout'] = {
+        'format': 'string',
+        'script_data': stdout
+    }
+
+    outputs['_stderr'] = {
+        'format': 'string',
+        'script_data': stderr
+    }
```
jeffbaumes: Will there also be a special way to map an input id to stdin for the container? Also, `_stdout` and `_stderr` should be able to be mapped to named output ids.

We will also want streaming output - we will want to look at the current stdout, stderr for a running docker container in the web interface. Monitoring construction/size/contents of the output files would be nice too, but may not need the level of inspection that stdout/stderr should support.

romanesco/tasks/docker.py
```diff
@@ -0,0 +1,101 @@
...
+
+
+def _expandArgs(args, inputs, taskInputs, tmpDir):
+    """
+    Expands arguments to the container execution if they reference input
+    data. For example, if an input has id=foo, then a container arg of the form
+    $input{foo} would be expanded to the runtime value of that input. If that
+    input is a filepath target, the file path will be transformed into the
+    location that it will be available inside the running container.
+    """
```
jeffbaumes: +1 seems like a good expansion. In the future we will also need conditional switches based on the value of a boolean input as we discussed, and possibly some other types of translation of inputs to command line args.

romanesco/tasks/docker.py
zachmullen: > Will there also be a special way to map an input id to stdin for the container?

I hadn't planned on adding that on this topic, but if it's a priority feature I can do so.

> Also, _stdout and _stderr should be able to be mapped to named output ids.

The way I've written it, stdout and stderr from the docker process are simply always included as special outputs. What sort of control do you envision for this? I could see only adding them as output if they are specified as outputs using a special key (perhaps `_stdout` and `_stderr`)?

> We will also want streaming output - we will want to look at the current stdout, stderr for a running docker container in the web interface. Monitoring construction/size/contents of the output files would be nice too, but may not need the level of inspection that stdout/stderr should support.

In the case of the COVALIC worker, the stdout of the docker container is actually the output rather than a side effect -- maybe that's bad, but it's one case we do want to support. I haven't yet implemented files as output, but I agree with your points in general. We'll need controls for specifying what purpose stdout and stderr are being used for, and ideally we could use them for streaming signaling like progress, messages, etc, and have most docker tasks write their output as files underneath /data.

jeffbaumes: No stdin is not a priority feature - just brainstorming.

I suppose since we now have name and id as separate fields in outputs, we could rely on the id to match e.g. `"_stdout"` to map directly to the output stream - so I think that's a good approach. I wouldn't want `_stdout` to appear in the user interface, but it will be hidden due to the separate UI-friendly name field.

It is a bit tricky how to interpret stdout. I suppose if an output is mapped to stdout then we just capture and stream stderr. If nothing is mapped to stdout we could stream both stdout and stderr to the client. Does that seem reasonable? I agree that it is convenient (and will be common practice) to just send the true algorithm output to stdout when using docker containers, so we certainly want to support that.

zachmullen: I like that idea: if `_stdout` or `_stderr` are explicitly specified as outputs, we'll save them in memory as outputs, and if not, we'll stream print them so they'll be streamed up to the job log.

To clarify, these would need to be specified in the task outputs, right?

jeffbaumes: Yes, they would need to be explicit task outputs with special ids. It may be a bit better to use a special `stream` flag instead of `_` - so instead of `"id": "_stdout"` you would name the `id` whatever you like and would have the special field `"stream": "stdout"` in your task output spec. So then you'd have

```python
    # Find output_id, the id of an output marked with "stream": "stdout" (if it exists). Then:
    outputs[output_id] = {
        'format': 'string',
        'script_data': stdout
    }
```

zachmullen:
The main thing missing from this is the ability to grab output files written by the container and push them as outputs. Do you want me to implement that on this topic, or should we do that on a separate one?

jeffbaumes:
If it's enough for COVALIC, I'd say merge this now and worry about that later. I am getting the following error when running `ctest -R docker -VV`, any idea the issue? I see that Travis doesn't mind so must be my setup:

```
test 4
    Start 4: docker

4: Test command: /usr/bin/python "-m" "unittest" "-v" "tests.docker_test"
4: Test timeout computed to be: 1500
4: Traceback (most recent call last):
4:   File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/runpy.py", line 162, in _run_module_as_main
4:     "__main__", fname, loader, pkg_name)
4:   File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/runpy.py", line 72, in _run_code
4:     exec code in run_globals
4:   File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/unittest/__main__.py", line 12, in <module>
4:     main(module=None)
4:   File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/unittest/main.py", line 94, in __init__
4:     self.parseArgs(argv)
4:   File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/unittest/main.py", line 149, in parseArgs
4:     self.createTests()
4:   File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/unittest/main.py", line 158, in createTests
4:     self.module)
4:   File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/unittest/loader.py", line 130, in loadTestsFromNames
4:     suites = [self.loadTestsFromName(name, module) for name in names]
4:   File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/unittest/loader.py", line 100, in loadTestsFromName
4:     parent, obj = obj, getattr(obj, part)
4: AttributeError: 'module' object has no attribute 'docker_test'
1/1 Test #4: docker ...........................***Failed    0.05 sec
```

zachmullen:
`pip install -r requirements-dev.txt`

For some reason unittest eats import errors while loading the test initially, we see the same thing in Girder and I haven't figured out a good way to fix it.

jeffbaumes:
Yup that did it. I had blindly tried `pip install -r requirements.txt` - did not see the `-dev` one sneak in there. LGTM!

PR #59, opened 2015-04-24
When creating a collection, add user permission to folders

zachmullen:
This fixes https://github.com/Kitware/tangelohub/issues/20

### DIFF
--- a/server/__init__.py
+++ b/server/__init__.py
@@ -128,9 +128,11 @@ def load(info):
             public=public, creator=user)
 
         for name in ('Data', 'Analyses', 'Visualizations'):
-            self.model('folder').createFolder(
+            folder = self.model('folder').createFolder(
                 name=name, public=public, parentType='collection',
                 parent=collection)
+            self.model('folder').setUserAccess(
+                folder, user=user, level=AccessType.ADMIN, save=True)
 
         return self.model('collection').filter(collection, user=user)
 


curtislisle:
Is it feasible to provide an interface at either the girder to TangeloHub levels to allow a private collection to be read/write for a few users? Is this already available, or is this an API level thing?  We want in Arbor to have a private-to-the-world, but shared-by-team collections.  How should this be accomplished?

zachmullen:
@curtislisle The permission management features on girder collections already enable this at both the REST API level and the girder web app level, but there is no GUI for this within tangelohub yet. We have all of the girder javascript widgets available inside the tangelohub app, so it wouldn't be too hard to throw in an access control widget for the collections that would facilitate granular permission management directly in tangelohub.

jeffbaumes:
The full Girder interface is available at https://arbor.kitware.com/girder which enables this (lock button when you browse to a collection/folder). I believe you currently need to add permission for each folder (data, analysis, visualization). I'll add an issue to support this directly in tangelohub as a one-step process if we want to add that later.

zachmullen:
Good point @jeffbaumes . I have been planning for a while to add a recursive permission setting checkbox to the access control widget, but haven't gotten around to it yet.

jeffbaumes:
Verified that a normal user cannot create an analysis on a newly-made collection on master, but it does work on this branch. LGTM!

PR #72, opened 2015-05-13
WIP: Add basic NetCDF support to romanesco

jeffbaumes:
For now there are two formats: `binary` is the blob of NetCDF (the contents of a `.nc` file) and `dataset` is an instance of the `Dataset` object in the `netCDF4` Python library. Validation on the `binary` format is minimal.

This needs testing and inspection by @aashish24 and @jbeezley.


romanesco/format/__init__.py
```diff
@@ -248,6 +248,6 @@ def import_default_converters():
     import_converters([os.path.join(cur_path, t) for t in [
         "r", "table", "tree",
         "string", "number", "image",
-        "boolean", "geometry"]])
+        "boolean", "geometry", "netcdf"]])
```
aashish24: Nice. In the future, I am assuming that we may need to have some mechanism for plugins? so that this list won't grow too big?

romanesco/format/netcdf/binary_to_dataset.json
```diff
@@ -0,0 +1,7 @@
+{
+    "name": "Binary to Dataset",
+    "inputs": [{"name": "input", "type": "netcdf", "format": "binary"}],
+    "outputs": [{"name": "output", "type": "netcdf", "format": "dataset"}],
```
aashish24: +1

aashish24: should we call it netcdf-dataset? dataset may be too generic but I am fine with calling it just dataset.

romanesco/format/netcdf/validate_binary.json
```diff
@@ -0,0 +1,6 @@
+{
+    "inputs": [{"name": "input", "type": "netcdf", "format": "binary"}],
+    "outputs": [{"name": "output", "type": "boolean", "format": "boolean"}],
```
aashish24: @jeffbaumes what does this mean? format": "boolean".

romanesco/format/netcdf/binary_to_dataset.json
jeffbaumes: Format is subsidiary to type - both are always needed. That is, in UI's, API's etc the full format will show as something like "netcdf:dataset". Other types may also have the format "dataset" without conflict.

romanesco/format/__init__.py
jeffbaumes: Yes there should be a dynamic way to define and import types, even external to the romanesco directory tree.

romanesco/format/netcdf/validate_binary.json
jeffbaumes: Everything needs a format, even in trivial cases like this. I've defaulted to using the same name as the type for the "natural" type. "inline" may also be a decent way to describe it.

For the "boolean" type (similar for "number") there are two formats: "boolean" and "json". The difference is:

```json
```

aashish24:
thanks @jeffbaumes. I thin this looks like a good start. I am merging it. Thanks for clarification.

PR #73, opened 2015-05-16
Add a new data type that uses python pickling

jbeezley:
This is provides a (nearly) generic serialization for python types.  Although pickles are somewhat error prone, this should enable us to generate some more generic tasks that take advantage python duck typing.

### DIFF
--- a/romanesco/format/__init__.py
+++ b/romanesco/format/__init__.py
@@ -248,6 +248,6 @@ def import_default_converters():
     import_converters([os.path.join(cur_path, t) for t in [
         "r", "table", "tree",
         "string", "number", "image",
-        "boolean", "geometry", "netcdf"]])
+        "boolean", "geometry", "netcdf", "pickle"]])
 
 import_default_converters()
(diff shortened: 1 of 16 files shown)


jeffbaumes:
Please rename `pickle` type name to `python`, with formats `pickle`, `pickle.base64`, and `object`.

aashish24:
@jbeezley just wanted to get more info on this. How does it help in the gaia/romanesco integration? I have some ideas but wanted to make sure I understand it.

jbeezley:
It allows (almost) arbitrary python objects to be used as data.  @jeffbaumes also mentioned this morning that there is a related type in the `R` data type as well.

aashish24:
thanks, I saw your other branch. Making more sense to me. I will review that code and will get back to you with more questions

jbeezley:
@jeffbaumes Assuming the tests pass this should be good to go.

jeffbaumes:
Yes this exactly what I was thinking. LGTM.

PR #74, opened 2015-05-17
Gaia task mode for romanesco

jbeezley:
Uses #73 to add a `gaia` task mode.  The new mode takes objects like this:
```python
spec = {
    "mode": "gaia",
    "script": "from gaia.core.operators import Add", # arbitrary setup code
    "class": "Add"                                   # the "Task" subclass to run
}
```

I'm a little skeptical about the usefulness of this over just writing the python script directly.  It does save some boilerplate code defining the input/output ports, but that is about all.

I have [another branch](https://github.com/kitware/romanesco/tree/gaia-compat) I'm working on that is probably more useful.  It does the opposite; allowing `gaia` to run `romanesco` style tasks.  Effectively, this wraps `romanesco` functions in an object oriented interface.


aashish24:
Nice.. looking into it now.

PR #76, opened 2015-05-23
Add shortcut from CSV to column names

jeffbaumes:
Does not require walking through the entire dataset

### DIFF
--- a/romanesco/format/table/csv_to_column_names.py
+++ b/romanesco/format/table/csv_to_column_names.py
@@ -1,5 +1,5 @@
-import cStringIO
+from six import StringIO
 from romanesco.format import csv_to_rows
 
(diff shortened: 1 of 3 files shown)


romanesco/format/table/csv_to_column_names.py
```diff
@@ -0,0 +1,5 @@
+import cStringIO
+from romanesco.format import csv_to_rows
+
+first_line = cStringIO.StringIO(input).readline()
```
zachmullen: use `six.StringIO` to be py2/3 compliant

jeffbaumes:
Added python 3 compatibility. Should be ready for final review.

zachmullen:
LGTM

PR #77, opened 2015-06-04
Implementing pythonic Spec API

jbeezley:
This implements (more or less) the API we discussed before[1].  In addition, this breaks backward compatibility in a couple of minor ways:
* Removed gaia package I'll add the relevant portions back as `gaia/gaia/` to package as a separate pip package later.
* Removed `romanesco/specs.py`.  I don't think it was being used and it's purpose is largely depreciated with the new API.
* Moved `romanesco/tasks/` to `romanesco/executors/`.  I think that is a better name, but I'm willing to move it back.

[1] https://gist.github.com/jbeezley/86dba3c13f424f36e8c1


zachmullen:
> Removed romanesco/specs.py. I don't think it was being used and it's purpose is largely depreciated with the new API.

:+1: , these were some stubs I wrote during the refactor and their functionality is superseded by your changes.

jbeezley:
I updated the [gist](https://gist.github.com/jbeezley/86dba3c13f424f36e8c1) to make it runnable with the current branch.  In the process, I found a bug that occurs with mutable default values (`PortList` objects namely).  I need to add a mechanism to generate defaults rather than store them.

aashish24:
:+1:

zachmullen:
Is this ready to merge?

jeffbaumes:
Yes LGTM - I think @aashish24's :+1: is a LGTM proxy

PR #82, opened 2015-06-09
[WIP] Add support for running Spark jobs in romanesco

cjh1:


romanesco/executors/pyspark.py
```diff
@@ -0,0 +1,35 @@
...
+    # Ensure that the spark modules have been added to our path
+    # Create SparkContext - Should we share one across a workflow?
+    # Take any inputs and convert them to RDDS?
+    print 'Running pyspark'
+
+    sc = spark.create_spark_context()
+
+    for name in inputs:
+        custom.__dict__[name] = inputs[name]["script_data"]
+        custom.__dict__['sc'] = sc
```
zachmullen: We may want to name this differently to avoid collisions with user-defined symbols. Maybe "spark_context" or "_spark_context"?

cjh1: The reason I chose "sc" is that is the variable the spark-shell would create for you, so we would then be somewhat compatible.

zachmullen: That's a good point. We should certainly pull it out of the loop body :) I suggest:

```
sc = spark.create_spark_context()
custom.__dict__['sc'] = sc
custom.__dict__['spark_context'] = sc  # alias in case user-defined input is named "sc"

for name in inputs:
    custom.__dict__[name] = inputs[name]["script_data"]

Comment (cjh1) [inline romanesco/executors/pyspark.py]: yup, we don't want to create one for each input!

File: romanesco/executors/pyspark.py [hunk 2]
```diff
@@ -0,0 +1,37 @@
...
+def run(task, inputs, outputs, task_inputs, task_outputs, **kwargs):
+    custom = imp.new_module("custom")
+
+    # TODO
+    # Ensure that the spark modules have been added to our path
+    # Create SparkContext - Should we share one across a workflow?
+    # Take any inputs and convert them to RDDS?
+    print 'Running pyspark'
+
+    sc = spark.create_spark_context()
```
zachmullen: A note on security, just so we are all on the same page: the spark context grants access to the SparkConf that was used to initialize it, which means anyone allowed to run scripts in this mode would be able to read all of the spark configuration values very easily: `print(sc._conf.getAll())`. There's not really an easy way around this -- if we wanted to hide it, we'd have to carefully wrap the spark context object to forbid access to the configuration.

Spark apparently only allows authentication via a per-application shared secret, so this is something we might want to consider.

romanesco/spark.py
```diff
@@ -0,0 +1,48 @@
...
+    except Exception as ex:
+        six.raise_from(Exception('Unable to create SparkContext, check Spark installation'), ex)
+
+
+def create_spark_context():
+    from pyspark import SparkConf, SparkContext
+    # Set can spark configuration parameter user has specified
+    spark_config = SparkConf();
+    for (name, value) in romanesco.config.items('spark'):
+        spark_config.set(name, value)
```
zachmullen: I like the simplicity of this behavior, but I think ideally we'd want to pass in the spark config parameters at runtime, at least optionally as an override to these fixed config settings. That way, a single romanesco worker group could be used to schedule jobs on multiple spark clusters and applications.

cjh1: Yes, I was planning on allowing that ( this is very much WIP :-) ). Settings such as spark.master definitely need to be past at runtime. Whats the "romanesco" way of doing this? Are they just extra inputs to the analysis?

romanesco/executors/pyspark.py
cjh1: Yep, I figured for the moment as we are allowing users to run arbitrary python code on the server it would be pretty hard to protect this config and in many case the user is going to setup this config.

romanesco/spark.py
zachmullen: Heh, sorry for the premature review. I think these should be passed as fields in the `task` argument when the mode is set to pyspark (or any other supported spark modes).

```
{
'mode': 'pyspark'
'spark_conf': {
    'spark.master': '...',
    'spark.app.name': '...',
    '<etc>...'
    }
}
```

cjh1: Not at all, my view is review early and often :-) I appreciate the feedback.

romanesco/__init__.py
```diff
@@ -19,9 +19,13 @@
     'docker': executors.docker.run,
     'python': executors.python.run,
     'r': executors.r.run,
-    'workflow': executors.workflow.run
+    'workflow': executors.workflow.run,
+    'pyspark': executors.pyspark.run
```
jeffbaumes: Interesting to name it `pyspark`. This leaves open a native `spark` mode which takes Scala code. Is that the idea here?

cjh1: Yes,  that was my thinking. However, may be it should be 'spark.python', 'spark.r', 'spark.scala' ... ?

zachmullen: :+1: for that name change

cjh1:
@jbeezley Any reason you closed this?

zachmullen:
This was auto-closed because the target branch was deleted. We should reopen against latest master.

jbeezley:
Yeah, didn't realize you had opened it against my branch.  You should be able to rebase on the latest master without conflicts.

cjh1:
Done

romanesco/__init__.py
jeffbaumes: I also like that name change.

cjh1: Good, I have updated it in the new PR.

requirements.txt
```diff
@@ -10,3 +10,4 @@ requests==2.6.0
 rpy2==2.3.10
 wsgiref==0.1.2
 six==1.9.0
+py4j==0.8.2.1
```
aashish24: minor thing.. just missing new line I guess

cjh1: Yep, fixed in latest PR, thank.

PR #83, opened 2015-06-09
Add support for running Spark jobs in romanesco

cjh1:


romanesco/__init__.py
```diff
@@ -19,9 +19,14 @@
     'docker': executors.docker.run,
     'python': executors.python.run,
     'r': executors.r.run,
-    'workflow': executors.workflow.run
+    'workflow': executors.workflow.run,
+    'spark.python': executors.pyspark.run
```
aashish24: +1

romanesco/executors/__init__.py
```diff
@@ -1 +1 @@
-from . import docker, python, r, workflow
+from . import docker, python, r, workflow, pyspark
```
aashish24: I guess we will need some sort of plugin in the future

zachmullen: I am happy to add some basic plugin infrastructure whenever we are ready to pull the trigger on that.

romanesco/spark.py
```diff
@@ -0,0 +1,55 @@
...
+
+    sys.path.append(os.path.join(spark_home, 'python'))
+    sys.path.append(os.path.join(spark_home, 'bin'))
+
+    # Check that we can import SparkContext
+    try:
+        from pyspark import SparkConf, SparkContext
+    except Exception as ex:
+        six.raise_from(Exception('Unable to create SparkContext, check Spark '
+                                 'installation'), ex)
```
zachmullen: When testing this out on my machine I ran into this and discovered that `six.raise_from` is sadly not very useful in python2 -- it just hides the `from` exception completely and raises the first argument. Maybe we should just remove the try/except block and let the error propagate up?

cjh1: Sure, I can do that or may be we should all move to python 3 :-)

zachmullen: Heh, I definitely want to support 2.7 and 3.3/3.4. The rest of romanesco will need to be sixified, but that shouldn't be too much work.

cjh1:
@zachmullen @jeffbaumes  @aashish24  I would like to get this reviewed and merge if possible. There is still more to do but I think this is a good spot to take stock ...

zachmullen:
This seems to work in my initial pass; I'm in the process of creating a toy girder plugin that exposes a REST endpoint for executing a spark job. I will share it as a tarball after lunch, and maybe we can use that as a base for running Chris's examples.

jbeezley:
Is there any way of mocking the pyspark calls to create some tests for this?  It barely passes the rather arbitrary limits I put in place for coverage decrease on [coveralls](https://coveralls.io/r/Kitware/romanesco).

aashish24:
I am good with the changes. A dummy test (something simple for now) would be nice as @jbeezley suggested.

cjh1:
@jbeezley I did add  some "real" spark tests. I can mock out the pyspark calls if sparks not available?

jbeezley:
It would be nice to have something that runs on travis.  I'm assuming spinning up a spark instance would be too difficult.

cjh1:
Not sure, might not be, would have to download the tarball and then run the master. Would be nice to have a "real" test.

jbeezley:
That could work, just remember, you can't use root and the list of packages you can install via apt is [limited](https://github.com/travis-ci/apt-package-whitelist/blob/master/ubuntu-precise).

cjh1:
@zachmullen Sounds good, I have converting the example to run in romanesco

romanesco/executors/pyspark.py
```diff
@@ -0,0 +1,30 @@
...
+import sys
+from romanesco import spark
+
+
+def run(task, inputs, outputs, task_inputs, task_outputs, **kwargs):
+    custom = imp.new_module("custom")
+
+    sc = kwargs['_romanesco_spark_context']
+    custom.__dict__['sc'] = sc
+    custom.__dict__['spark_context'] = sc
```
zachmullen: One other thing that might be handy to expose here is the spark module itself, since otherwise I imagine most everyone would have to include `import spark` in their script.

zachmullen: Forget that, I was just confused by the spark code examples :P

zachmullen:
As promised: https://github.com/zachmullen/sparktest

Right now it just points at a single hardcoded path as the input, but it's a starting point for fleshing out some more robust examples.

cjh1:
:sparkles is working on Travis :smile_cat:

jbeezley:
:+1:

tests/spark_test.py
```diff
@@ -0,0 +1,66 @@
...
+            analysis['script'] = script
+
+        outputs = romanesco.run(analysis, {
+                'a': {'format': 'json', 'data': '[1,2,3,4,5,6,7,8,9]'}
+            },
+            {
+                'b': {'format': 'json'}
+            })
+
+        expected = {'b': {'data': '[1, 4, 9, 16, 25, 36, 49, 64, 81]', 'format': 'json'}}
```
aashish24: I may be wrong here but isn't it b has format spark.rdd?

cjh1: This is the data binding and in this case I am testing the conversion from spark.rdd ( the output of the analysis ) to json. So this is right.

jbeezley: It gets converted automagically because of line 59 above.

jbeezley: And, I'm 13 seconds too slow.

aashish24: that's what I thought but just wanted to make sure.  So if b was not listed with format json then we would have gotten spark.rdd as well?

aashish24:
LGTM :+1:

tests/spark_test.py
cjh1: Correct

aashish24:
Ship it!

cjh1:
@jeffbaumes @zachmullen @jbeezley Are you guys happy for me the merge this?

jeffbaumes:
LGTM

zachmullen:
LGTM

jbeezley:
[LGTM](http://lgtm.in/i/UyXVZo0dq)

cjh1:
@jbeezley That is awesome!

cpatrick:
I came here to do what @jbeezley did... LGTM!

jbeezley:
We clearly need to use http://www.lgtm.in/ more.

aashish24:
I think @jbeezley beat you by few minutes :smile:

PR #85, opened 2015-06-09
Add some more formalized docs for the romanesco.run dictionary specs

zachmullen:


jbeezley:
Excellent.  :clap:

docs/api-docs.rst
```diff
@@ -1,6 +1,217 @@
...
+square braces, which represent lists in python or Arrays in JSON.
+
+.. code-block :: none
+
+    <TASK> ::= <PYTHON_TASK> | <R_TASK> | <DOCKER_TASK> | <WORKFLOW_TASK>
+
+    <PYTHON_TASK> ::= {
+        "mode": "python",
+        "script": <python code to run as a string>
+        (, "inputs": [<TASK_INPUT> (, <TASK_INPUT, ...)])
```
jeffbaumes: Need closing bracket. Same on lines 41, 49, 57.

docs/api-docs.rst
```diff
@@ -1,6 +1,217 @@
...
+***********************
+
+The ``inputs`` argument to :py:func:`romanesco.run` specifies the inputs to the
+task described by the ``task`` argument. Specifically, it tells what data should
+be placed into the task input ports.
+
+.. code-block:: none
+
+    <INPUTS> ::= {
+        <id> : <INPUT_BINDING>
```
jeffbaumes: Should probably be `[<id>: <INPUT_BINDING> (, <id> : <INPUT_BINDING>, ...)]` to show there can be multiple inputs.

docs/api-docs.rst
```diff
@@ -1,6 +1,217 @@
...
+
+.. code-block :: none
+
+    <TASK> ::= <PYTHON_TASK> | <R_TASK> | <DOCKER_TASK> | <WORKFLOW_TASK>
+
+    <PYTHON_TASK> ::= {
+        "mode": "python",
+        "script": <python code to run as a string>
+        (, "inputs": [<TASK_INPUT> (, <TASK_INPUT>, ...)])
+        (, "outputs": [<TASK_OUTPUT> (, <TASK_OUTPUT>, ...)])
```
jeffbaumes: This would be a lot cleaner if JSON allowed for trailing commas, but I like the precision here that following this exactly will result in a spec in proper JSON form.

docs/api-docs.rst
```diff
@@ -1,6 +1,217 @@
...
+    <INPUT_BINDING> ::= <INPUT_BINDING_HTTP> | <INPUT_BINDING_LOCAL> |
+                        <INPUT_BINDING_MONGODB> | <INPUT_BINDING_INLINE>
+
+    <INPUT_BINDING_HTTP> ::= {
+        "mode": "http",
+        "url": <url of data to download>
+        (, "headers": <dict of HTTP headers to send when fetching>)
+        (, "method": <http method to use, default is "GET">)
+        (, "maxSize": <integer, max size of download in bytes>)
+    }
```
jeffbaumes: I believe all input bindings should have formats. I see the `format` field for output bindings but not input bindings.

docs/api-docs.rst
```diff
@@ -1,6 +1,217 @@
...
+The mongodb input mode specifies that the data should be fetched from a mongo
+collection. This simply binds the entire BSON-encoded collection to the input
+variable.
+
+.. code-block:: none
+
+    <INPUT_BINDING_INLINE> ::= {
+        "mode": "inline",
+        "data": <data to bind to the variable>
+    }
```
jeffbaumes: I believe this is (or should be) the default mode if no mode is provided, so we should document that.

jeffbaumes:
LGTM when my simple comments are addressed. Great job with this.

docs/api-docs.rst
zachmullen: Just looked through the code, it is true that the format is required when `auto_convert=True`, so I am going to document it as a required field.

docs/api-docs.rst
zachmullen: We don't currently have a default input binding mode, the logic looks like this:

```
elif mode == 'inline':
        return spec['data']
    else:
        raise Exception('Unknown input fetch mode: ' + mode)
```

Should I open an issue to change that behavior?

jbeezley: It probably needs some clarification.  There is different behavior depending on if you call `convert` or `fetch` as I recall.  One of those works with no `mode`.

jbeezley: See this bit of code:

https://github.com/Kitware/romanesco/blob/master/romanesco/__init__.py#L192

jeffbaumes: Oh I didn't realize this was explicit - I think it will help with shorthand of inlined parameters like numbers and strings to not require the mode to be set, so yes I'd suggest opening an issue for that. You can keep the docs as is until that is changed.

PR #91, opened 2015-06-17
Allow entrypoint override in docker exec mode

zachmullen:


zachmullen:
Just noticed the coveralls output here in github... that's neat :)

jbeezley:
All the cool kids are doing it.

zachmullen:
How hard is that to get going? I wouldn't mind doing it on girder if it's easy.

jbeezley:
For pure python, it is basically just [this](https://github.com/Kitware/romanesco/blob/master/.travis.yml#L67).  To include the javascript coverage as well, you need export in a compatible format.  Last I looked, I don't think it will work with the cobertura files we are currently creating.  I believe blanket does support an output mode that is supported though.  My knowledge is about 4 months old by now, so everything could be different.

As a bonus, coveralls calculates the total coverage correctly as opposed to cdash which just averages all the numbers together.

cpatrick:
It would be great to teach CTest to output coverage in an appropriate format for coveralls. It would be similarly great to teach CDash to ingest that coverage format as well.

@jbeezley Could you explain more about what CDash gets incorrect?

tests/docker_test.py
```diff
@@ -122,3 +122,16 @@ def fetchMock(url, request):
             self.assertRegexpMatches(cmd2[6], _tmp + '/.*:/data')
             self.assertEqual(cmd2[7:],
                              ['test/test:latest', '-f', '/data/file.txt'])
+
+            # Make sure we can specify a custom entrypoint to the container
+            mockPopen.reset_mock()
+            task['entrypoint'] = '/bin/bash'
```
jeffbaumes: Can you update our shiny new spec documentation with this optional field?

zachmullen:
Knew I forgot something @jeffbaumes . Fixed :)

jeffbaumes:
LGTM on test passage.

PR #93, opened 2015-06-24
Fixes #92. Add a "directory" type, and a couple formats

### LINKED ISSUE #92: New type and formats: directory/zip, directory/tgz
I've been trying to think of a good way to fit the process of unarchiving an input file into the romanesco architecture, and I think this way makes sense. Example:

```
"task": {
    "inputs": [{
        "id": "a_directory",
        "description": "This is the path to the unarchived directory",
        "type": "string",
        "format": "string"
    }]
}

"inputs": {
    "a_directory": {
        "type": "directory",
        "format": "tgz",   // or "zip"
        "target": "filepath",
        "mode": "http",
        "url": "https://data.kitware.com/file/foo/download" // location of the archive
    }
}
```

Then we can simply rely on the auto-convert process to do the unzipping.

@jeffbaumes does that make sense?

zachmullen:
This was added for the purpose of supporting automatic extraction
of zip and tar.gz archives into directories using just the type
and format fields.


jeffbaumes:
LGTM. Eventually would be good to have the reverse process but (dir to tar/zip) but we can leave that until a use case arises for it.

zachmullen:
Agreed. I will open another issue for that.

PR #95, opened 2015-06-30
Move settings from config file into setting model

zachmullen:


jeffbaumes:
LGTM, very nice! I was able to test tangelohub running while changing various settings without restarting Girder - all worked as expected to give and revoke access.

PR #102, opened 2015-07-14
Remove open collection creation endpoint

zachmullen:
This functionality will be controllable in girder core settings.


zachmullen:
The second commit (fc6e671) adds the new setting I suggested that will allow non-admin users to dynamically control access to task execution.

jeffbaumes:
LGTM along with https://github.com/Kitware/tangelohub/pull/154

PR #104, opened 2015-07-21
Factor out code that runs a romanesco task

cjh1:
To allow this code to be called from other Girder plugins without
having to make a RESTful call.


zachmullen:
LGTM

PR #106, opened 2015-07-29
Plugin system

zachmullen:


zachmullen:
@jeffbaumes @jbeezley @cjh1 this is probably a good place to cut this branch off for review. I am going to open up future work as issues.

scripts/install_requirements.py
```diff
@@ -0,0 +1,65 @@
...
+import subprocess
+import sys
+
+
+def installReqs(file, pip):
+    if not os.path.exists(file):
+        return
+
+    print('\033[32m*** Installing: %s\033[0m' % file)
+    if subprocess.call([pip, 'install', '-U', '-r', file]) != 0:
```
jbeezley: I think this would be safer (at least as a default) given multiple python installs:
```python
import pip
pip.main(['install', '-U', '-r', file])
```

I can't think of a reason you would use a pip not in the current `PYTHONPATH`.

jbeezley:
Aside from my one comment, it looks good to me.  Might want a second opinion though.

cjh1:
There are obviously a lot of changes here so its hard to review/understand everything, but the spark stuff LTGM and I look forward to removing the R dependency in minerva :-)

jeffbaumes:
LGTM! Nice work @zachmullen

aashish24:
:clap:  It would be nice if we can document the plugin system

aashish24:
Taking my words back. Looks like you did add some documentation.

PR #109, opened 2015-08-10
Adding a Graph Type

danlamanna:


jeffbaumes:
Taking a look at this now, I think I'd change the name of the `graph` format to `networkx`. This is more explicit about the nature of the format.

zachmullen:
Should we place this into a plugin since it has new external package requirements?

jeffbaumes:
Excellent question @zachmullen. Currently table and tree are considered core, and having graph in core would seem to round out the standard general-purpose data structures. networkx is apparently a simple and quick install so I lean toward having it in core. Of course we could go the other way and have very little in core, including making a tree plugin. But for how things are setup now I lean toward having graph in core.

zachmullen:
Sounds good!

romanesco/format/graph/adjacencylist_to_networkx.py
```diff
@@ -0,0 +1,5 @@
+from networkx.readwrite.adjlist import read_adjlist
+from StringIO import StringIO
```
zachmullen: To avoid regressions when we eventually add python3 support, we should change this to `from six import StringIO`.

danlamanna: Makes sense, I noticed occurrences in other places so should I go ahead and do a project-wide replace for it?

danlamanna:
I've added all of the vtkGraph converters with the exception of vtkGraph -> NetworkX.

This is because it seems as though the methods of retrieving the data from a vtkTree or vtkTable are not the same as vtkGraph (and not readily apparent).

Unlike vtkTree, ```vtkGraph.GetVertexData.GetNumberOfArrays``` always yield 0 for me, and ```GetVertices``` requires an iterator which when given, goes on to infinity. (This could just be my shallow understanding of VTK)

@jeffbaumes - I dug into the vtk bindings a little bit, but am unsure how much time you want me spending on investigating scraping the data out of a vtkGraph further.


The other concern I have is with NetworkX -> vtkGraph conversion:
I'm [currently assuming](https://github.com/Kitware/romanesco/blob/3b1883b5777c1275c60e5c90d6dafedec3a6f761/romanesco/plugins/vtk/converters/graph/networkx_to_vtkgraph.py#L9-L14) that metadata associated with nodes and edges are associated with ALL nodes and edges and are of the same type.

This seems to be the same approach that [tree/nested -> vtkTree](https://github.com/Kitware/romanesco/blob/3b1883b5777c1275c60e5c90d6dafedec3a6f761/romanesco/plugins/vtk/converters/tree/nested_to_vtktree.py#L7-L11) is taking - but I think in the case of graphs we may want something more robust. Thoughts?

curtislisle:
Hi Dan,
Thanks for working on the graph type into romanesco! In response to the attribute question, I would like to register my desire to support asymmetric attributes (where nodes don't all have the same attributes). We can certainly force symmetric attributes if we have to, but a lot of future graphs will have different numbers of attributes, annotations, etc. on nodes.   The networkx code I've written just copies all attributes it discovers. We could force symmetric if exporting to vtkGraph instances.

jeffbaumes:
Thanks for the comments @curtislisle. @danlamanna is working to support asymmetric attributes for all formats where it is feasible. The vtkGraph conversions will not have symmetric attributes for now.

danlamanna:
I think this is ready for (at least initial) review.

@jeffbaumes Should we add Roni to the repo to have him take a look at 3d9aa88?

ronichoudhury:
This looks good, Dan.  The only concern I have is, these won't work well
for "big data"-scale graphs, since the conversion scripts both depend on
holding the entire input/output in memory.  This is one reason it's easier
to work with Mongo's semi-JSON format of one document per line.  If there
is a way to "stream in" elements of a JSON list programmatically (via
special purpose iterator, I guess), that would solve this concern as well.

jeffbaumes:
@ronichoudhury I agree this is certainly an issue for large graphs. This branch is mostly for a baseline for the graph type, akin to what we had for table and tree up until we started doing things with Spark. A follow-on for @manthey and @danlamanna (and you) to think about is what are the big data ways to deal with graphs and how would we link them to romanesco.

One simple approach is to store a pointer to a dataset (like a mongo collection) in a string and pass that around as a string type. @curtislisle has proposed that approach as an initial way to work with large graphs. The analysis internals would do any streaming needed by iterating through the database. I can also see adding more official integration into large-graph analytic frameworks where there would be an specific graph format representing a pointer to a graph in that framework.

jeffbaumes:
This is looking good but @ronichoudhury can you do a final review since I'll be on vacation after today.

jeffbaumes:
Also @danlamanna please edit PR title to remove WIP if this is ready for final review.

ronichoudhury:
No problem.  I'll wait for the [WIP] to disappear before I do it.

romanesco/format/graph/networkx_to_clique_json.py
```diff
@@ -0,0 +1,42 @@
...
+output = []
+
+# There is no clearly defined notion of undirected graphs in clique.json
+# It simply checks for edges going in either direction, so we ignore
+# whether it is an nx.DiGraph or just nx.Graph
+
+for (node, data) in input.nodes(data=True):
+    clique_node = {
+        '_id': {
+            '$oid': node
```
ronichoudhury: Using ``node`` as the OID won't work in general, because Mongo expects a 24-character hex string for this OID value.  The simplest way to make this right is to hash ``node`` to generate an appropriate string and use that as the OID.  This makes the fixup to lines 29 and 32 below simpler (``u`` and ``v`` will just be replaced with the same hashing strategy used on ``node``).  Converting from Clique to NetworkX would simply copy over the OID to serve as the node label.

This means that the ``node`` label on the NetworkX graph will be ephemeral but this seems to be an acceptable tradeoff, at least for now.

(Dan and I just discussed this in person; leaving this note here for everyone else.)

ronichoudhury:
Aside from [this](https://github.com/Kitware/romanesco/pull/109#discussion_r37110438), LGTM.

danlamanna:
@ronichoudhury 

I decided to go the route of generating a random ```ObjectId``` and mapping it - this still achieves the same outcome, we're just not hashing ```node``` to generate the id. Since ```NetworkX``` allows any python hashable object to be a ```node```, md5 hashing ```[1,2,3]``` wouldn't necessarily work, it would have to be coerced to a string, and so on.

ronichoudhury:
LGTM!

jeffbaumes:
Late to the party, but can't you store anything as `_id`? It doesn't have to be an `ObjectId` for Mongo's sake, so it could just be an ID, index, etc. from the original graph. But perhaps Clique assumes/requires `ObjectId`?

ronichoudhury:
Clique only assumes uniqueness of the "key" field in general, but it might expect the Mongo backend to contain ``ObjectId``s in the ``_id`` field (which are subsequently used as "keys" for the nodes).  I can change this behavior in Clique if it's considered too restrictive.

Issue #111, opened 2015-08-11
Use more networkx builtins


danlamanna:
Right now we [build a large object in memory](https://github.com/Kitware/romanesco/blob/803d3def59203cf06289b5e88ab449c0f3201f6e/romanesco/format/__init__.py#L62), and re-compute it every time a plugin adds a validator/converter.

With ```networkx``` being part of core soon (see #109) - we could potentially refactor/remove [a bit](https://github.com/Kitware/romanesco/blob/803d3def59203cf06289b5e88ab449c0f3201f6e/romanesco/format/__init__.py#L130-L149) of [code](https://github.com/Kitware/romanesco/blob/803d3def59203cf06289b5e88ab449c0f3201f6e/romanesco/utils.py#L211-L246) to use networkx's builtins like [topological sort](https://networkx.github.io/documentation/latest/reference/generated/networkx.algorithms.dag.topological_sort.html) by building up our workflows/validators/converters as DiGraphs and then finding conversion paths using something like [single_source_dijkstra](https://networkx.github.io/documentation/latest/reference/generated/networkx.algorithms.shortest_paths.weighted.single_source_dijkstra.html#networkx.algorithms.shortest_paths.weighted.single_source_dijkstra) (I mention this one because it also has a cut-off in  number of hops away which we currently implement).

This would still involve storing a large object in memory but would only compute paths when they needed to be used, thus allowing plugins to add validators/converters as nodes/edges without any additional computation.

zackgalbreath:
I'll defer to @jeffbaumes on this, but this seems like a good idea to me.  The less roll-your-own code we have to maintain the better.  It would be compelling if we could do a before/after comparison to show that this change results in improved performance and/or reduced memory usage.

zachmullen:
:+1: 

We also have our own toposort implementation that we could get rid of if we want.

jeffbaumes:
@zachmullen I don't think performance/memory will be measurable in the graph sizes we are working with (a few dozen converters so a few dozen links). The biggest potential improvement here is a decrease in our code complexity. Getting rid of a 5-level `for` loop sounds like a win. :+1: 

I like the idea of doing a shortest path search at runtime as long as it's stable (will give the same path each time). I think we could get rid of the arbitrary cutoff of path length 3 and allow arbitrary lengths. I opened #112 as a way for the user to know what happened in the conversion process, including the number of conversions used.

zachmullen:
[...]

PR #113, opened 2015-08-14
NetworkX Integration

danlamanna:
~~This is dependant on #109 being merged into master~~


danlamanna:
From c90af08 (first commit on this branch not on graph-type):
> Re #111

> This introduces the use of validators and converters as a NetworkX
DiGraph instead of a nested dictionary.

> Type/Formats are represented as a 2 element tuple such as ('string',
'text')

> The nodes of the DiGraph contain the analysis needed to validate an
input of it's own format. The edges contain the analysis needed to
convert an input going from its source validator to its target
validator.

> This doesn't break any backwards compatibility (to my
knowledge). Plugins can continue to call import_converters on certain
paths and they will just be added to the DiGraph. Conversion paths are
computed at the moment they need to be used.

> This removes the notion of a 3-hop cutoff. This is something that will
be noted in the future by introducing an "explain"-like operation which
will show the path different analyses will have to take.

danlamanna:
This is ready for review.

I was also going to refactor [this section](https://github.com/Kitware/romanesco/blob/ce710fab022ff1c9d5c58bc227035792c1243bd5/romanesco/executors/workflow.py#L9-L60) of the code to use ```NetworkX ```, but after starting to get into it it didn't seem like it would cut down on many lines of code or make it much clearer. Would appreciate thoughts on that though.

danlamanna:
What are thoughts on using a [namedtuple](https://docs.python.org/2/library/collections.html#collections.namedtuple) to represent a Type/Format validator, and passing that around?

Such as:
```Validator(type='graph', format='networkx')```   
I only suggest it because I've already started passing around tuples in [some cases](https://github.com/Kitware/romanesco/commit/ce710fab022ff1c9d5c58bc227035792c1243bd5#diff-de45e0045dbec440bfe15fe351c1fb54R63), and multiple arguments in [others](https://github.com/Kitware/romanesco/commit/ce710fab022ff1c9d5c58bc227035792c1243bd5#diff-de45e0045dbec440bfe15fe351c1fb54R81).

**Comment by curtislisle** [comment]
I like having several different named subtypes of graph and being able to identify them from each other.  This seems like a good way to accomplish it.

**Comment by danlamanna** [comment]
@zachmullen Can you review this? I would be interested to hear your thoughts on my ```namedtuple``` remark as well.

**Comment by zachmullen** [comment]
> I would be interested to hear your thoughts on my namedtuple remark as well.

Whatever we do, we should be consistent about it. A namedtuple would be preferable to a tuple, or we could just keep using two kwargs. Keeping things named is good for readability.

File: romanesco/format/__init__.py [hunk 1]
```diff
@@ -54,17 +59,97 @@ def csv_to_rows(input):
...
+def converter_path(source, target):
+    """Gives the shortest path that should be taken to go from a source
+    type/format to a target type/format.
+
+    Throws a NetworkXNoPath exception if it can not find a path.
+
+    Returns a list of edges in the order they should be traversed.
+    """
+    # These are to ensure an exception gets thrown if source/target don't exist
+    get_validator(*source)
```
zachmullen: We should make the specification of source and target consistent between the functions so we don't need to use `*`. I'm fine with your namedtuple suggestion, but if we go with that, let's make all of the functions expect that type, and we should document it with sphinx conventions for said functions.

romanesco/format/__init__.py
```diff
@@ -99,53 +181,38 @@ def import_converters(search_paths):
...
-                    input_format[out_format] = {}
-                input_format[out_format] = [analysis]
+        return analysis
 
-    os.chdir(prevdir)
+    prevdir = os.getcwd()
+    for path in search_paths:
+        os.chdir(path)
+        validator_files = set(glob.glob(os.path.join(path, "validate_*.json")))
+        converter_files = set(glob.glob(os.path.join(path, "*.json"))) - validator_files
```
zachmullen: I think a better convention would be to change `*.json` to `*_to_*.json, and still subtract `validator_files` on the off chance some format name contains `_to_`.

tests/format_test.py
```diff
@@ -0,0 +1,44 @@
...
+        self.assertFalse(has_converter('string', 'text', 'number'))
+
+        # There is a converter from string/text -> string/json
+        self.assertTrue(has_converter('string', 'text', 'string', 'json'))
+
+        # There are no self loops in the conversion graph
+        self.assertFalse(has_converter('string', 'text', 'string', 'text'))
+
+        # Converters don't go from one type to another
+        self.assertFalse(has_converter('string', out_type='number'))
```
zachmullen: Great test. One are we don't cover is the [printing functions for the graph](http://my.cdash.org/viewCoverageFile.php?buildid=811015&fileid=1308227), `print_conversion_graph` and `print_conversion_table`. Could you add some basic testing for those in here as well by reassigning `sys.stdout`?

danlamanna:
I had to technically make ```Validator``` a class to support sphinx style docs.. see [this SO post](http://stackoverflow.com/a/13785187).

danlamanna:
Rebased to incorporate new plugin changes.

danlamanna:
Rebased on master for #123 so travis is happy, and added 77c1ff6.

Shortest path can be stable in ```NetworkX``` by using the ```OrderedDict``` graph implementations but the issue is if plugins were loaded in different orders then they could produce a different conversion path. Finding all shortest paths (which will usually be 1-2), sorting, and picking the first will always produce a stable conversion path relative to which plugins are loaded (regardless of the order they were loaded in).

romanesco/format/__init__.py
```diff
@@ -54,17 +74,109 @@ def csv_to_rows(input):
...
+    get_validator(target)
+
+    # We sort and pick the first of the shortest paths just to produce a stable
+    # conversion path. This is stable in regards to which plugins are loaded at the
+    # time.
+    paths = all_shortest_paths(conv_graph, source, target)
+    path = sorted(paths)[0]
+    path = zip(path[:-1], path[1:])
+
+    return [get_edge(u, v)[2] for (u, v) in path]
```
zachmullen: Why do we need to call get_edge here? Don't we already have the edges within `path`?

danlamanna: We have the to and from ```Validator``` objects but not the associated analysis, that is stored in the edge data. ```NetworkX``` edges are formatted as ```(u, v, data)``` so ```get_edge(u, v)[2]``` is just returning the analysis. I've cleaned this up and documented it in d36400a

romanesco/format/__init__.py
```diff
@@ -54,17 +74,113 @@ def csv_to_rows(input):
...
+
+        if target.type and target.type != v.type:
+            continue
+
+        if target.format and target.format != v.format:
+            continue
+
+        return True
+
+    return False
```
zachmullen: This function seems to only look for direct connections rather than doing a full graph traversal. If that is intentional, we should document that fact, since otherwise someone might think this simply checks if the conversion is possible at all.

danlamanna: Phew, thanks. This was a bug. cfbd1e6 adds a test to catch it, and fixes it.

danlamanna:
@zachmullen This is set (again).

danlamanna:
Rebased, fixed conflicts with ce865a03

kotfic:
+1 for this btw

zachmullen:
This LGTM, sorry for the slow review.

I vote that once this and #118 go in, we should cut a release.

jeffbaumes:
Got rid of two triple for loops and one quintuple for loop - nice job fixing several of the more complex parts of the codebase!

Issue #110, opened 2015-08-11
Allow for plugins to have their own tests


danlamanna:
Right now, although VTK functionality is [in the plugins](https://github.com/Kitware/romanesco/blob/803d3def59203cf06289b5e88ab449c0f3201f6e/romanesco/plugins/vtk/__init__.py) - the testing for it remains in the core of romanesco.

@jeffbaumes and I thought it might be good for plugins to have the ability to contain their own tests, and cmake/ctest would be able to pick up on it.

PR #116, opened 2015-08-18
Encapsulate tests for plugins

danlamanna:
Re #110


zachmullen:
Thanks, I didn't get around to supporting this in my initial branch :+1:

danlamanna:
@zachmullen If you like how 93d260b implements it, then I'll go ahead and port the rest of the plugin-specific tests over.

I want to make sure that's done right, specifically in regards to coverage, and if we're ok with
```add_python_test(graph PLUGIN vtk PLUGINS_ENABLED vtk)``` as opposed to having CMake pick up on what directory/plugin it is and the user leaving ```PLUGIN vtk``` out.

zachmullen:
:+1:

danlamanna:
5eed106 highlights an inconvenience with having to rename tests because they may exist in a plugin as well as core.

Additionally, I left the table/tree plugin tests in core because some of them use converters from the ```vtk``` plugin *and* the ```r``` plugin, so they are codependent but it doesn't make sense to group them together.

danlamanna:
Ready for review.

zachmullen:
This LGTM

PR #117, opened 2015-08-19
Girder io plugin

zachmullen:


aashish24:
:clap: @kotfic are you going to review this branch?

aashish24:
I guess @mgrauer  as well

mgrauer:
I'm definitely not the right person to review this, as I'm not a Romanesco developer and have no context for this PR.

:rabbit:

zachmullen:
Context, for anyone who is interested: simple HTTP IO works well enough for fetching data from girder, but for actually pushing data, it's not sufficient since Girder requires a more complicated upload process than just a single request. So, we created a plugin that leverages the awesome power of `girder_client` to make this possible. This also makes downloading directories from girder more straightforward, as they won't need to be archived by the server and un-archived in the task.

mgrauer:
Sounds like a great idea.

kotfic:
LGTM,  thanks!

PR #118, opened 2015-08-19
Executor debugging

kotfic:
Adds code, tests, documentation for 'debug' runtime flag which can be set on an python of spark.python analysis,  or passed to ```romanesco.run()```.  In the case of the analysis the task script will be written to file before executed.  In the case of being passed to ```romanesco.run()``` all tasks (including validators and converters) will be written to file.


kotfic:
@zachmullen PTAL - thanks!

zachmullen:
My only hesitation about this is that we may want to come up with a better name than `debug`; we may want to use the word "debug" to mean something different in the future, and this is a fairly special-purpose behavior. I would prefer something like `write_script`.

danlamanna:
This looks good to me, flake8 found a few things but it's currently only being used for ```/specs```. I removed the unused imports.

danlamanna:
Before adding 2 blocks of identical code, is this maybe a good time to abstract the executor? Seeing as how the pyspark executor is [essentially the same thing](https://gist.github.com/danlamanna/30766e697783d1f49106).

zachmullen:
@danlamanna can you also make it so we are running flake8 on the whole romanesco package rather than just `specs`? I think that's just some typo that made it in.

kotfic:
@zachmullen I've changed from 'debug'  to 'write_script'  for the flag.  @danlamanna  Executor refactor seems out of scope for the  branch,  maybe open an issue?

zachmullen:
LGTM.

PR #124, opened 2015-08-21
[WIP] - AST based python parser for Gaia/Romanesco

kotfic:
This commit adds a prototype AnalysisStaticParser class that Parses
files with romanesco python and spark.python tasks using abstract syntax
trees. The returned parse is valid romanesco analysis JSON. This is done
by wrapping injected variables in metaclasses that derive from
RomanescoTypeConverter. In a script this looks like:
```python
from classes import Table, Output

def main(a, b):
    c = {'fields': a['fields'], 'rows': a['rows'] + b['rows']}
    return c

if __name__ == "custom":
    c = Output(main(Table(a, format="rows"), Table(b, format="rows")),
               type="table", format="rows")
```
Where Table() and Output() are defined in classes and main() is expected
to do some kind of computation on a and b. Classes inheriting from
RomanescoTypeConverter, when executed,  currently return the variable
unchanged.  If the above script were placed into a file "test_analysis.py" one
could call the following code:

```python
AnalysisStaticParser("test_analysis.py", "python").parse()
```
And it would return the following JSON:
```Javascript
{'inputs': [{'format': 'rows', 'name': 'a', 'type': 'table'},
            {'format': 'rows', 'name': 'b', 'type': 'table'}],
 'mode': 'python',
 'name': 'test_analysis',
 'outputs': [{'format': 'rows', 'name': 'c', 'type': 'table'}],
 'script': '...omitted...'}
```
To see this in action you can run:
```
cd examples/
python run_example.py
```


kotfic:
@jbeezley  This is a prototype of what we were talking about yesterday.  if you could take a look and maybe we could chat about how this overlaps/doesn't overlap with the spec and port  stuff?

@zachmullen Some of the things I am doing here are a little esoteric (e.g.,  metaclasses, abstract syntax trees)  If you could comment on possible gotcha's that i'm missing i would appreciate it.  

@aashish24 This is my impression of the way in which you want Gaia to head,  if you could give this a quick review in terms of larger goals of gaia that would be great

romanesco/plugins/gaia/classes.py
```diff
@@ -0,0 +1,160 @@
...
+
+module = sys.modules[__name__]
+
+for key, val in validators.items():
+    romanesco_types.append(key.title())
+
+    if not hasattr(module, key.title()):
+        setattr(module,
+                str(key.title()),
+                type(str(key.title()), (RomanescoTypeConverter, ), {}))
```
zachmullen: Could we make `RomanescoTypeConverter` a metaclass that registers its subclasses in `romanesco_types` for us? Also couldn't `romanesco_types` be made into a dict rather than a list so we could just store all the dynamic classes within the dict rather than within `module.__dict__`?

(Also the loop header could just be rewritten `for key in validators.keys():`)

jbeezley: +1 to both of those suggestions.  I attempted something similar in the old gaia for registering subclasses.  https://github.com/OpenGeoscience/gaia/blob/master/gaia/core/factory.py

jbeezley:
This doesn't really overlap with anything I had done.  There should probably be some integration between this and the [port](https://github.com/Kitware/romanesco/blob/master/romanesco/specs/port.py) spec.  It's not terribly clear to me at the moment what that integration should look like though.

aashish24:
Thanks @kotfic. I am not clear on certain things here. Would you mind if we talk more tomorrow on this?

romanesco/plugins/gaia/classes.py
kotfic: @zachmullen @jbeezley moving ```romanesco_types``` into ```RomanescoTypeConverter``` makes sense to me. 

The goal of putting those classes into ```module.__dict__``` was to make them directly importable from the module allowing end users to wrap injected variables in things like ```Table()```  or ```String()```   or ```Graph()``` classes.  Unless i'm missing something-  If we push the classes inside ```RomanescoTypeConverter```  then it becomes more complicated to wrap injected variables (e.g.,  ```RomanescoTypeCoverter.romanesco_type["string"]()```) right?

My hope was this would be a half measure between a full OOP refactor of the Type/Format system and that eventually the metaclass stuff could be cleared out of this plugin -  or is that not in the cards for romanesco?

kotfic:
@aashish24 Sure thing,

In addition to the AST parser,  i'd like to propose an API for Gaia workflows that looks like this:

```python
import networkx as nx

# PythonTask would wrap the AST parser, could also
# Imagine PySparkTask()  or RTask() or ScalaTask()
t1 = PyTask("/path/to/add.py")   #  => 
t2 = PyTask("/path/to/multiply.py")
t3 = PyTask("/path/to/multiply.py")

# Define the workflow graph using NetworkX,  input
# and outputs are set using edge metadata and validated
# under the covers. 
task_graph = nx.DiGraph()
task_graph.add_edges_from([(t1, t3, {"b": "in1"}) 
                           (t2, t3, {"b": "in2"})])

# Define a workflow
wf = Workflow(task_graph)

# Could be used to generate the actual workflow JSON
wf.parse() # => "{ 'mode': 'workflow',
           #       'inputs': [...],
           #       'outputs': [...],
           #       'steps': [...],
           #       'connections': [...] }"

# Or run the workflow as a romanesco analysis using port/spec 
# interface behind the scenes
if __name__ == "custom":
    x, y = 2, 4
    output = wf(Number(x), Number(y)) # => romanesco.run('{.....}', mode="worflow", ...)


```

Under the covers the ```PyTask``` and ```Workflow``` classes can manage the port/spec integration and provide a  consistent pythonic interface to the the task specs.  Comments?

romanesco/plugins/gaia/classes.py
jbeezley: I see what you are going for.  My primary concern was with name collisions in the module.  I just realized you are already checking for that.  I would maybe issue a warning or something on the `else` here in that case.

kotfic:
This comment is to capture notes from a meeting between @kotfic and @jeffbaumes about some useful conceptual distinctions when talking about romanesco analysis script data,  and meta data.  The general thrust of the conversation was (I believe)  how can we make it easier for people to write romanesco tasks?  Pros and Cons lists are just cursory stabs - additional suggestions are welcome.

### Python inside Metadata
#### An example of the current implementation
Currently an analysis definition maintains two separate files,  one for script data,  and one for meta data.

_script data from file csv_to_column_names.py:_
```python
from six import StringIO
from romanesco.format import csv_to_rows

first_line = StringIO(input).readline()
output = csv_to_rows(first_line)['fields']
```
_metadata from file csv_to_column_names.json:_
```javascript
{
    "name": "CSV to Column Names",
    "inputs": [{"name": "input", "type": "table", "format": "csv"}],
    "outputs": [{"name": "output", "type": "table", "format": "column.names"}],
    "script_uri": "file://csv_to_column_names.py",
    "mode": "python"
}
```
The two files exist so that we can edit one in python,  and the other in javascript,  but _really_  the ```file://csv_to_column_names.py``` file is loaded at run time and so within romanesco the analysis definition is a python dict:

```python
{
    "name": "CSV to Column Names",
    "inputs": [{"name": "input", "type": "table", "format": "csv"}],
    "outputs": [{"name": "output", "type": "table", "format": "column.names"}],
    "script": "from six import StringIO\nfrom romanesco.format import csv_to_rows\n\nfirst_line = StringIO(input).readline()\noutput = csv_to_rows(first_line)['fields']\n"
    "mode": "python"
}
```
I will categorize this as a "**Python inside Metadata**" solution.   What jeff and I talked about were several "**Metadata inside Python**" solutions. I would argue that the AST parser is a "**Python ___as___ Metadata**" solution.

### Metadata inside Python solutions:
One solution is to require a function that produces the necessary metadata for running that task.  This function could be consistently named,  or could be decorated in some way to mark it as the describing function. Essentially the function would take no arguments,  and would return the necessary metadata.

#### Function endpoint
```python
from six import StringIO
from romanesco.format import csv_to_rows

def describe():
    return {
        "name": "CSV to Column Names",
        "inputs": [{"name": "input", "type": "table", "format": "csv"}],
        "outputs": [{"name": "output", "type": "table", "format": "column.names"}],
        "mode": "python"}

first_line = StringIO(input).readline()
output = csv_to_rows(first_line)['fields']
``` 

This could be hand coded as in the above example,  or could use the Spec/Port functionality  for a more "pythonic" solution:

```python
from six import StringIO
from romanesco.format import csv_to_rows
from romanesco import Spec, Port

def describe():
    spec = Spec(mode="python")
    spec.add_input(Port({"name": "input", "type": "table", "format": "csv"}))
    spec.add_output(Port({"name": "output", "type": "table", "format": "column.names"}))
    return spec

first_line = StringIO(input).readline()
output = csv_to_rows(first_line)['fields']
``` 
- *Pros:*
  - No dependencies
  - Single file solution
- *Cons:*
  - Totally explicit
  - Potentially non-trivial amount of extra code

#### Decorators
Another approach would be to use decorators attached to some module level object. e.g.:

```python
from six import StringIO
from romanesco.format import csv_to_rows
from romanesco import Task

task = Task()

@task.input(name="input", type="table", format="csv")
@task.output(type="table", format="rows")
def run(a, b):
    first_line = StringIO(input).readline()
    return csv_to_rows(first_line)['fields']
   
```
This would allow you to maintain input and output state in the ```task``` variable and generate a metadata dict in romanesco with the ```run``` function in an executable code object

- *Pros*:
  - No injected variables
  - Clean pythonic interface
- *Cons*:
  - Issues with specifying multiple outputs
  - Requires all scripts to be written as a known or decorated function

### Python as Metadata solution:
The Abstract Syntax Tree (AST)  solution prototyped in this branch is a third solution that integrates the Python and the metadata into a single file.  It leverages the parsing process to pull metadata out of the file,  and the exec process to run the script.  Following the same example:

```python
from six import StringIO
from romanesco.format import csv_to_rows
from romanesco import String, Output

first_line = StringIO(Table(input, format="csv")).readline()
output = Output(csv_to_rows(first_line)['fields'], type="table", format="rows")
```

Here we have wrapped the ```input``` variable in a function ```Table(..., format="csv")``` and wrapped the rvalue of the assignment operator in an ```Output(..., type="table", format="rows")``` function.

When being *executed*  both ```Table()``` and ```Output()``` are the identity function. They pass back what was passed into them.   However *before*  execution we can use an AST parser to pull this data out of the file. 

- *Pros*:
  - Integrated script and metadata
  - Supports multiple outputs
  - Marks variables that should be injected
  - In the current implementation these functions are just the identity function - but the _could_ do other things,  like validate or convert the injected variables.
- *Cons*:
  - AST parser _may_ be difficult to maintain (weird edge cases?)
  - Undefined variables are still present in the file (Though see: https://github.com/Kitware/romanesco/pull/138 for a possible convention that makes the file command line/module friendly)

cpatrick:
Personally, I prefer something more like decorators or an explicit specification (passing a dict to a function in the code) over the AST solution. Not necessarily the use of the AST, but the "magic" appearance of the way it is implemented here doesn't feel very Pythonic (i.e. explicit is better than implicit).

kotfic:
@cpatrick actually i was hoping this made this *more* explicit.  Currently we have undefined variables that are injected into the code at run time.  These can be be anywhere in the file and are completely unmarked aside from the accompanying JSON file.  The ```Table``` and ```Output``` functions at least visually mark where these variables are in the file.

cpatrick:
@kotfic Well, we certainly should be more specific than we are now :smile_cat:. I guess I put in my vote for the function call approach.

kotfic:
I have removed the AST parsing code from this branch and will continue with gaia plugin code here.  I am also renaming the pull request.   I still feel AST parsing is a good solution to the Python/Metadata problem.  Even so,  it doesn't belong in this branch which was originally intended to implement geospatial datatypes and tasks.    I will keep the discussion/comments around for historical purposes unless someone can think of a better way to archive them.

jbeezley:
I would just close this PR and create a new one with the relevant commits cherry picked.

danlamanna:
I'm going to put this here even though it's closed just to record it.

My vote is for the AST - I think it makes writing analyses much easier, and works as a complete drop in replacement without breaking any existing code or analyses. I find it to actually be pretty explicit, with the exception of the auto-generation of type classes like ```Table``` and so on, but I don't believe that is necessarily set in stone.

I'm not very familar with running romanesco workers, but from our meeting it seemed that this is the only approach that completely sidesteps the issue of having dependencies on the server and worker side of things, since all it does is actually parse the code rather than execute it (avoiding imports).

My next vote would be for an object oriented approach, something similar to this [VisTrails example](http://www.vistrails.org/usersguide/v2.2/html/packages.html#an-example-module) with an interface/abstract class enforcing a ```describe``` function exist.

aashish24:
> My next vote would be for an object oriented approach, something similar to this VisTrails example with an interface/abstract class enforcing a describe function exist.

Yup, that's what (similar) we are doing after for Gaia (we were discussing earlier).

PR #131, opened 2015-08-26
Test environment should not take plugins enabled value from cfg file

zachmullen:
This was breaking some tests in development environments where plugins
were enabled.


danlamanna:
LGTM

PR #133, opened 2015-08-27
Make Configuration one level lower in docs

jeffbaumes:
Without this Sphinx makes a top-level item on the left that does not highlight when clicked. I think there is only supposed to be one top-level Sphinx heading per page.


danlamanna:
LGTM

PR #139, opened 2015-08-28
Conversion Graph Cleanup

danlamanna:
From f5de399:
> Removing O(n) calls over NetworkX edges/nodes in favor of O(1) calls
through their built-in data structures.

> Removing node metadata that was never actually being used.

> Simplifies get_validator so it just returns the runnable analysis.

Also fixed a couple of instances where I wasn't using ```six```.

### DIFF
--- a/romanesco/format/__init__.py
+++ b/romanesco/format/__init__.py
@@ -86,19 +86,9 @@ def converter_path(source, target):
     :returns: An ordered list of the analyses that need to be run to convert from
     ``source`` to ``target``.
     """
-    def get_edge_analysis(source, target):
-        """Return the analysis associated with an edge in the conversion graph,
-        given the source Validator and the target Validator.
-        """
-        for (u, v, data) in conv_graph.edges([source], data=True):
-            if v == target:
-                return data
-
-        return {}
-
     # These are to ensure an exception gets thrown if source/target don't exist
-    get_validator(source)
-    get_validator(target)
+    get_validator_analysis(source)
+    get_validator_analysis(target)
 
     # We sort and pick the first of the shortest paths just to produce a stable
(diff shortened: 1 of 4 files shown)


romanesco/format/__init__.py
```diff
@@ -86,19 +86,9 @@ def converter_path(source, target):
...
-            if v == target:
-                return data
-
-        return {}
-
     # These are to ensure an exception gets thrown if source/target don't exist
-    get_validator(source)
-    get_validator(target)
+    get_validator_analysis(source)
+    get_validator_analysis(target)
```
kotfic: maybe consider wrapping these in a try/catch  and throwing a NetworkXNoPath exception instead of the generic Exception()  thrown by get_validator_analysis ?

romanesco/format/__init__.py
```diff
@@ -222,14 +206,10 @@ def get_analysis(filename):
...
-            in_format = analysis["inputs"][0]["format"]
 
-            conv_graph.add_node(Validator(in_type, in_format), {
-                "type": in_type,
-                "format": in_format,
-                "validator": analysis
-            })
+            conv_graph.add_node(Validator(analysis["inputs"][0]["type"],
+                                          analysis["inputs"][0]["format"]),
+                                analysis)
```
kotfic: Would it be nicer if we had a more consistent API for this -  e.g.,  ```get_input_type(analysis)```,  ```get_input_format(analysis)```  same for outputs?

danlamanna: Validators are a special case only ever taking one input, and only ever returning one output. I'm not sure how we would handle that type of API for multiple inputs/outputs analyses in general.

Regardless I'm adding a comment noting the magic numbers.

kotfic:
LGTM

PR #140, opened 2015-08-31
[WIP] - Gaia romanesco plugin

kotfic:
Closed https://github.com/Kitware/romanesco/pull/124,  opening this PR to track conversation around gaia romaneco plugin (minus AST parsing)


kotfic:
Stubbed out implementation of the workflow API

```python
import GeopandasReader, GeopandasWriter
import networkx as nx

class GeopandasReader(Task):
    __inputs__ = ["..."]
    __outputs__ = ["..."]

    def __init__(self, *args, **kwargs):
        # Init Stuff
        pass

    def parse(self):
        return {
            "inputs": ["..."],
            "outputs": ["..."]
            "..."
        }
    
    def run(self):
        # run code
        return [out for out in self.__outputs__]


class Workflow(Task):
    __inputs__ = ["..."]
    __outputs__ = ["..."]
    
    def __init__(self,  graph):
        self.graph = graph

    def parse(self):
        self.tasks = [n.parse for n in self.graph.nodes()]
        self.connections = [n.parse for n in self.graph.edges()]
            
        return {
            "tasks": self.tasks,
            "connections": self.connections,
            "..."
        }
    def run(self, inputs, outputs):
        romanesco.run(self.parse(),  inputs, outputs)


if __name__ == "__romanesco__":
    t1 = GeopandasReader(uri="file:///path/to/shapefile.pth",
                         format="ESRI Shapefile")
    t2 = GeopandasWriter(uri="file:///path/to/geojson.geojson",
                         format="GeoJSON")
    g = nx.Digraph()

    g.add_edge((t1, t2))

    wf = Workflow(g,  connection_params={})

    wf.parse() # =>  {....}
    
    wf.run()
```

romanesco/plugins/gaia/classes.py
```diff
@@ -0,0 +1,98 @@
...
+        ret = []
+
+        for pos, port in enumerate(self.__outputs__):
+            try:
+                ret.append(raw[port.name]['data'])
+            except KeyError:
+                # Error here?
+                pass
+
+        return ret[0] if len(ret) == 1 else tuple(ret)
```
kotfic: This ```Task``` object should probably be the task object from the spec.  I would like to see the two merged but only if we decide that that this is how we want to move forward (especially with class-level variables that define the ports on an analysis).

kotfic:
@jeffbaumes @aashish24 This is a prototype of what I believe Aashish is looking for in terms of the readers and writers.  It does not implement the workflow pseudo-code at the top of the PR but I would like too do that - assuming everyone agrees this is the right way forward.

romanesco/plugins/gaia/requirements.txt
```diff
@@ -0,0 +1,11 @@
...
+Shapely==1.5.11
+descartes==1.0.1
+cligj==0.3.0
+click==5.1
+click-plugins==1.0.1
+Fiona==1.6.1
+python-dateutil==2.4.2
+pandas==0.16.2
+pyproj==1.9.4
+geopandas==0.1.1
```
jbeezley: At some point we may want to cache the python path on travis because these packages take a while to install.

jbeezley:
+1 on merging the `Task` classes.  I would like to see some plugin docs added, but besides that it looks good to me.

aashish24:
Mostly looks good to me but let's wait before we can discuss this in the meeting today. Also, I am assuming that if we have the ability to run arbitrary run function then the we get rid of the script API?

romanesco/plugins/gaia/classes.py
```diff
@@ -0,0 +1,101 @@
...
+             type="string",
+             format="text",
+             validate=False)])
+    __outputs__ = PortList([Port(name="output",
+                                 type="boolean",
+                                 format="boolean")])
+
+    @property
+    def script(self):
+        return """
```
aashish24: can we not use python inspect module? for example inspect.getsource and inspect.getsourcelines will be very useful in getting function definition to string conversion. I know that they work in the interactive mode and will work for our use-case here since the definition is visible.

kotfic:
This has been primarily superseded by #147  Closing

PR #141, opened 2015-09-02
Visualization Tutorial/Example

danlamanna:
@jeffbaumes Here's a rough draft of the tutorial we spoke about.

I ran into a few issues using port/spec functionality, namely:
- Using a direct callable instead of a script tag bypasses romanesco.run which means auto-conversion isn't currently supported in that manner
- ```romanesco.load``` doesn't appear to be integrated into Task yet, so ```script_uri``` within a Task specification won't be fetched
- I ran into issues using Ports because romanesco.run expects the inputs to exist on the task dictionary as well.

I didn't spend too much time looking into them because I'm not sure where we're headed with this functionality - but I'm sure they could be resolved without a large amount of effort.


jeffbaumes:
This is looking very good. You should be able to build a workflow with JSON based on the two analysis specs, instead of running them separately. Look at the JSON spec documentation for a WORKFLOW_TASK to see how to do it.

danlamanna:
@jeffbaumes I updated this to use the WORKFLOW builtins, local URL to view it still works.

One thing I noticed was auto-conversion only exists on input, not on output. We could skip the ```convert``` step at the bottom if we specified the workflow output as ```networkx.json``` and it auto-converted.

jeffbaumes:
LGTM after you remove the call to `convert` as we discussed.

docs/examples.rst
```diff
@@ -0,0 +1,168 @@
...
+       475     484
+
+Each integer represents an anonymized Facebook user. Users belonging to the same line in the adjacency list indicates a symmetric relationship in our undirected graph.
+
+Build a Workflow
+~~~~~~~~~~~~~~~~
+
+Create a file named ``workflow.py``, this is the file we'll be using to create our Romanesco workflow.
+
+Finding the Most Popular Person
```
ronichoudhury: s/Finding/Find/

docs/examples.rst
```diff
@@ -0,0 +1,168 @@
...
+
+.. literalinclude:: static/facebook-example.py
+    :caption: workflow.py/most_popular_task
+    :lines: 1-22
+    :linenos:
+
+.. note :: As we saw with our last script assuming ``G`` would in be scope, this task explicitly states that both ``most_popular_person`` and ``G`` will be in scope (as its outputs) when it's done.
+
+
+Finding Their Neighborhood
```
ronichoudhury: s/Finding/Find/

docs/examples.rst
```diff
@@ -0,0 +1,168 @@
...
+
+.. literalinclude:: static/facebook-example.py
+    :caption: workflow.py/find_neighborhood_task
+    :lines: 24-44
+    :lineno-start: 27
+    :linenos:
+
+
+Putting It Together
+###################
```
ronichoudhury: s/Putting/Put/

ronichoudhury:
This looks great!

PR #144, opened 2015-09-07
Minimal swift integration

jeffbaumes:
Creates “swift” mode for http://swift-lang.org/ (not Apple’s swift). Support basic file output and basic string argument-based input, which seems to be a main use case.

This only works local, there are .conf files that can be used to setup swift on supercomputers that we would want to support.

### DIFF
--- a/romanesco/plugins/swift/executor.py
+++ b/romanesco/plugins/swift/executor.py
@@ -4,30 +4,11 @@ import romanesco.utils
 import tempfile
 
 
-def _transform_path(inputs, taskInputs, inputId, tmpDir):
-    """
-    If the input specified by inputId is a filepath target, we transform it to
-    its absolute path within the docker container (underneath /data).
-    """
-    for ti in taskInputs.itervalues():
-        tiId = ti['id'] if 'id' in ti else ti['name']
-        if tiId == inputId:
-            if ti.get('target') == 'filepath':
-                rel = os.path.relpath(inputs[inputId]['script_data'], tmpDir)
-                return os.path.join('/data', rel)
-            else:
-                return inputs[inputId]['script_data']
-
-    raise Exception('No task input found with id = ' + inputId)
-
-
 def _expand_args(args, inputs, taskInputs, tmpDir):
     """
     Expands arguments to the container execution if they reference input
     data. For example, if an input has id=foo, then a container arg of the form
-    $input{foo} would be expanded to the runtime value of that input. If that
-    input is a filepath target, the file path will be transformed into the
-    location that it will be available inside the running container.
+    $input{foo} would be expanded to the runtime value of that input.
     """
     newArgs = []
     regex = re.compile(r'\$input\{([^}]+)\}')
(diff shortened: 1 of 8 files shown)


romanesco/plugins/swift/executor.py
```diff
@@ -0,0 +1,112 @@
+import os
+import re
+import select
+import subprocess
+import sys
+import tempfile
+
```
jeffbaumes: @zachmullen I essentially lifted code from this file verbatim from your `docker` plugin. Should there be a (possibly core) shell execution mode where `docker` and `swift` (among others) utilize components of it to customize it?

zachmullen: It would be good to pull out the stuff inside `_run_swift_process` since it's the exact same as the docker one, and is generally useful for anything running the task as a subprocess. We could put that in the `utils` module in core perhaps.

jeffbaumes:
Ready for final review. @zachmullen moved run_process to utils. Is the way I changed docker test ok? I had to change how you mocked some functions since the `select` module import moved.

romanesco/plugins/swift/executor.py
```diff
@@ -0,0 +1,78 @@
...
+            if ti.get('target') == 'filepath':
+                rel = os.path.relpath(inputs[inputId]['script_data'], tmpDir)
+                return os.path.join('/data', rel)
+            else:
+                return inputs[inputId]['script_data']
+
+    raise Exception('No task input found with id = ' + inputId)
+
+
+def _expand_args(args, inputs, taskInputs, tmpDir):
```
jeffbaumes: @zachmullen should this be a utils function too?

romanesco/plugins/swift/executor.py
```diff
@@ -0,0 +1,78 @@
+import os
+import re
+import romanesco.utils
+import tempfile
+
+
+def _transform_path(inputs, taskInputs, inputId, tmpDir):
```
jeffbaumes: @zachmullen: utils function?

aashish24:
This is great!!. will help us with some other projects as well.

romanesco/plugins/swift/executor.py
zachmullen: Yep, this syntax will be good for any command line execution mode.

romanesco/plugins/swift/executor.py
zachmullen: I think this is specific to docker since it's mapping paths on the local filesystem to paths within the docker container.

romanesco/plugins/swift/executor.py
jeffbaumes: Since this calls `_transform_path` I'm tempted to leave this in both places for now - the swift one will not call `_transform_path` but the docker one will. I'll remove `_transform_path` from the swift plugin. Otherwise either (1) a transform function would need to be passed in (identity function for swift) or (2) path transforms would be a pre- or post-processing step, not embedded in `_expand_args`. I lean toward (2) as a pre-processing step, but don't feel like it's worth messing with at this point.

romanesco/plugins/swift/executor.py
jeffbaumes: Agreed.

romanesco/plugins/swift/executor.py
zachmullen: I also lean toward (2) :smile:

jeffbaumes:
@zachmullen ok I think the code is cleaner now with docker-specific stuff removed from the swift execution mode.

zachmullen:
LGTM, can we open an issue for the _expand_args refactor and link it to this PR?

jeffbaumes:
Sure @zachmullen will do.

PR #148, opened 2015-09-10
Allow includes of JS/CSS on read the docs

danlamanna:
I can't assure this will work (aside from it working on my local machine). Is there a way to test this branch on RTD?

Though, extending ```layout.html``` is currently broken it seems. This is an [open issue](https://github.com/rtfd/readthedocs.org/issues/152) on RTD. 

This, unfortunately means that occurrences in [tangelo](https://github.com/Kitware/tangelo/blob/develop/docs/templates/layout.html) are probably not working either.

### DIFF
--- a/docs/templates/page.html
+++ b/docs/templates/page.html
@@ -1,4 +1,4 @@
-{% extends "!layout.html" %}
+{%- extends "layout.html" %}
 
 {% block extrahead %}
   {% if pagename == 'examples' %}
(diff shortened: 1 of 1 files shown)


ronichoudhury:
What specific differences does this cause between your local build and the
RTD build?

danlamanna:
Before this commit, RTD is not loading the extrahead block which allows the SVG to show up at the bottom of my example.

With this commit, everything stays the same on my local build - I am hoping that it will fix the RTD build because it seems as though RTD ignores layout.html during the build process.

ronichoudhury:
Interesting.  I can't remember what, if anything, Tangelo needs from
layout.html.

danlamanna:
@jeffbaumes Are we okay with merging this? The only way to know if it will actually work on RTD is to try, unless we can build readthedocs off of a specific branch.

jeffbaumes:
Yeah let's merge it and see. If it doesn't work and we feel we need to iterate more we can open a new branch and figure out how to make RTD build that branch.

PR #150, opened 2015-09-10
Fix NetworkX Keyword Arguments

danlamanna:
This isn't strictly needed in Graph/DiGraph but it should be specified in MultiGraph/MultiDiGraphs as Roni and I discovered. 

Adding it to all instances of ```add_edge```... I suppose explicit *is* better than implicit.


ronichoudhury:
LGTM

PR #151, opened 2015-09-10
Fixing absolute path issue

danlamanna:
@jeffbaumes This should work, but I've said that before.


jeffbaumes:
LGTM

jeffbaumes:
Looks like it's alive, nice.

PR #154, opened 2015-09-15
Fixes #87. Allow "inline" mode to be default if data is specified

### LINKED ISSUE #87: Make "inline" the default IO binding mode

zachmullen:


jeffbaumes:
This works as I would expect, LGTM.

PR #155, opened 2015-09-16
Fix travis build error

jbeezley:
There is some problem with the tls 1.2 support for travis that I couldn't get around.  This moves the R and arbor packages to midas and sets them up manually.


kotfic:
LGTM

PR #156, opened 2015-09-16
Add is_valid function to Validator class

kotfic:
This implements an is_valid() function on the Validator class providing an API for validating Port specs.  

Previously ```has_converter``` was being used to validate type/format named tuples.  Unfortunately if a type/format exists but does not have a converter than ```has_converter``` will return False incorrectly. For example ```{type="netcdf",  format="dataset"}``` is a valid data type,  but has no converter (There is only a ```binary``` to ```dataset``` converter) exists. 

This makes validation of port metadata a separate distinct function.


jeffbaumes:
LGTM

PR #161, opened 2015-09-17
Girder io update file

kotfic:
This alters girder_io's behavior to use the ```uploadFile``` function instead of the ```uploadFileToItem``` function.  The primary difference is ```uploadFile``` can take a ```ByteIO``` object while ```uploadFileToItem``` expects a path to a file on disk. This allows the output of a romanesco task/workflow to be directly uploaded to the server,  rather than written to disk then re-read and uploaded.   This means the 'data' field of the output data-spec should be the actual data and not a path which is a breaking change.

### DIFF
--- a/romanesco/plugins/girder_io/__init__.py
+++ b/romanesco/plugins/girder_io/__init__.py
@@ -1,7 +1,7 @@
 import girder_client
+import six
 import os
 import romanesco
-
 
 def _init_client(spec, require_token=False):
     if 'host' not in spec:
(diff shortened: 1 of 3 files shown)


kotfic:
Forgot to note that ```uploadFile``` is in girder-client master,  but not in the current release.  This will require a girder-client release before it can be merged.

zachmullen:
@kotfic you should now be able to update to girder_client==1.1.0, it is now in pypi.

zachmullen:
Rather than making this an either/or proposition, we should support both. We already do this in the [HTTP IO mode](https://github.com/Kitware/romanesco/blob/master/romanesco/io/http.py#L76-L84), where there is an additional field on the spec called `target`, which is either set to `filepath`, indicating the data is a path to a file on disk that should be uploaded, or `memory`, indicating that this is something in memory, as you've written.

PR #162, opened 2015-09-17
Fix sphinx docstring warnings and add spec docs

jbeezley:
The api docs could probably be better organized with a bit of an introduction for each component, but at least includes them.

### DIFF
--- a/docs/types-and-formats.rst
+++ b/docs/types-and-formats.rst
@@ -91,8 +91,7 @@ A list of rows with ordered, named column attributes. Formats:
     ``"objectlist"`` format. This is the format of MongoDB collections.
 
 :``"csv"``: A string containing the contents of a comma-separated CSV file.
-    Column headers will be reasonably detected if present, otherwise
-    columns will be named ``"Column 1"``, ``Column 2"``, etc.
+    The first line of the file is assumed to contain column headers.
 
 :``"tsv"``: A string containing the contents of a tab-separated TSV file.
     Column headers are detected the same as for the ``"csv"`` format.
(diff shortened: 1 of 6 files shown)


danlamanna:
LGTM.

(I didn't get any docstring warnings before, running ```Sphinx==1.3.1```)

jbeezley:
You have to blow out your _build directory because the docstrings don't ever seem to get rebuilt.  I always just do `make clean html`.

docs/types-and-formats.rst
```diff
@@ -93,7 +93,6 @@ A list of rows with ordered, named column attributes. Formats:
 :``"csv"``: A string containing the contents of a comma-separated CSV file.
     Column headers will be reasonably detected if present, otherwise
     columns will be named ``"Column 1"``, ``Column 2"``, etc.
-    See `has_header`_ for details on header detection.
```
jbeezley: @jeffbaumes I removed this line because the link is invalid, and I couldn't make any sense of it.

danlamanna:
Reproduced, this fixes it. #justShipIt

docs/types-and-formats.rst
jeffbaumes: Yeah that's old stuff. I believe column headers are now assumed - so lines 94-95 should be removed as well.

docs/types-and-formats.rst
```diff
@@ -91,9 +91,7 @@ A list of rows with ordered, named column attributes. Formats:
     ``"objectlist"`` format. This is the format of MongoDB collections.
 
 :``"csv"``: A string containing the contents of a comma-separated CSV file.
-    Column headers will be reasonably detected if present, otherwise
-    columns will be named ``"Column 1"``, ``Column 2"``, etc.
-    See `has_header`_ for details on header detection.
+    The first line of the file is assumed to contain column headers.
```
jbeezley: @jeffbaumes How is this?

jeffbaumes: Excellent, thanks.

PR #164, opened 2015-09-21
Add validator endpoint to REST API

danlamanna:
This adds the capability to query Romanesco for the validators it has loaded, optionally filtering by the type and format.

### DIFF
--- a/server/__init__.py
+++ b/server/__init__.py
@@ -364,7 +364,7 @@ def load(info):
         (':itemId', 'romanesco', ':jobId'),
         romanescoStopRun)
 
-    info['apiRoot'].validator = Validator(getCeleryApp())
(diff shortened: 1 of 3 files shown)


server/__init__.py
```diff
@@ -363,5 +364,7 @@ def romanescoStopRun(jobId, params):
         (':itemId', 'romanesco', ':jobId'),
         romanescoStopRun)
 
+    info['apiRoot'].validator = Validator(getCeleryApp())
```
zachmullen: I think we should namespace this better, e.g. `romanesco_validator`.

zachmullen:
This change has some places that need PEP8 compliance, hopefully the failing tests will show those places.

danlamanna:
@zachmullen They should be fixed now.

zachmullen:
LGTM

PR #166, opened 2015-09-23
Fix cmake binary url on travis

jbeezley:


zachmullen:
Looks like wget is the same as curl as far as following redirects by default :P

zachmullen:
```
>>> r = requests.get('http://psg.mtu.edu/pub/apache/spark/spark-1.3.1/spark-1.3.1-bin-hadoop2.4.tgz', allow_redirects=False)
>>> r.status_code
302
>>> r.text
u'<!DOCTYPE HTML PUBLIC "-//IETF//DTD HTML 2.0//EN">\n<html><head>\n<title>302 Found</title>\n</head><body>\n<h1>Found</h1>\n<p>The document has moved <a href="http://psg.mtu.edu/fail.html">here</a>.</p>\n</body></html>\n'
>>> r.headers
{'date': 'Wed, 23 Sep 2015 14:36:39 GMT', 'content-length': '212', 'content-type': 'text/html; charset=iso-8859-1', 'location': 'http://psg.mtu.edu/fail.html', 'server': 'Apache'}
```

jbeezley:
The spark download is also failing now.  When it rains, it pours.  Maybe a temporary failure.  But +1 on the failure image there.

fail

zachmullen:
Heh, just noticed the "fail.html". Guess we can try again later.

jbeezley:
I updated the spark download location on this as well.  The new url is hosted by apache and so is probably more official.

jbeezley:
Does anyone want to give this a +1?

zachmullen:
Lgtm

PR #169, opened 2015-09-24
Image processing example

jbeezley:
This is what I came with to introduce tasks and workflows with a "real" example.  The workflow is a bit contrived; if anyone has any better ideas, let me know.  This example even has embedded doctests ensuring they continue working.  It is probably worth considering doing something similar to the facebook graph example as well.

Also, the build will fail until #166 is pulled in.


docs/facebook-example.rst
```diff
@@ -0,0 +1,161 @@
...
+
+   <div id="popularity-graph"></div>
+
+
+This is of course a more verbose than necessary workflow for the purposes of demonstration. This could have easily been done with one task,
+however by following this you should have learned how to do the following with Romanesco:
+
+ * Create Romanesco tasks which consume and produce multiple inputs and outputs
+ * Run Romanesco tasks as part of a multi-step workflow
+ * Use Romanescos converter system to serialize it in a format JavaScript can read
```
zachmullen: Romanesco's

docs/facebook-example.rst
```diff
@@ -0,0 +1,161 @@
...
+
+Put it together
+###############
+Conceptually, this is what our workflow will look like:
+
+.. image:: static/facebook-example-workflow.png
+    :align: center
+    :alt: Visualize Facebook Data Workflow Diagram
+
+\* The format changes because of Romanescos auto-conversion functionality.
```
zachmullen: Romanesco's

docs/facebook-example.rst
```diff
@@ -0,0 +1,161 @@
...
+using `d3.js <http://d3js.org/>`_.
+
+
+In this example we will:
+ 1. Obtain a set of Facebook data
+ 2. Find the most "popular" person in our data
+ 3. Find the subgraph of the most popular person's neighborhood
+ 4. Visualize this neighborhood using d3
+
+Obtain thie dataset
```
danlamanna: Mistake when correcting my mistake.

docs/image-example.rst
```diff
@@ -0,0 +1,253 @@
+Image processing
+----------------
+
+This example will introduce how to use Romanesco to build some simple image
+processing tasks using Pillow_.  We will learn how to chain several tasks
```
zachmullen: Somewhat pedantic, but I think we should just call this PIL or Python Imaging Library, since Pillow is just the distribution and that's a detail that users won't really see.

docs/facebook-example.rst
jbeezley: Obtain thy dataset

zachmullen: Thine available datasets are NORTH, EAST, and DOWN. :dragon_face:

docs/image-example.rst
```diff
@@ -0,0 +1,253 @@
...
+      'inputs': [
+         {'name': 'blur_input', 'type': 'image', 'format': 'pil'},
+         {'name': 'blur_radius', 'type': 'number', 'format': 'number'}
+      ],
+      'outputs': [{'name': 'blur_output', 'type': 'image', 'format': 'pil'}],
+      'script': '''
+   from PIL import ImageFilter
+   blur_output = blur_input.filter(ImageFilter.GaussianBlur(blur_radius))
+   '''
+   }
```
jeffbaumes: Do our Spec classes aid us at all in setting up these input bindings and tasks to make the code more compressed?

docs/image-example.rst
```diff
@@ -0,0 +1,253 @@
...
+   --- finished: blur ---
+   --- beginning: show ---
+   --- finished: show ---
+
+.. doctest::
+   :hide:
+
+   >>> isinstance(output['the_image']['data'], Image)
+   True
+
```
jeffbaumes: Would it be hard to embed the blurred image here in the docs?

danlamanna:
I really like the idea of testing our tutorials, I thought about it but didn't know [sphinx.ext.doctest](http://sphinx-doc.org/ext/doctest.html) existed. I'll add tests to mine afterwards.

jeffbaumes:
Good idea for testing your example @danlamanna. I really like these examples - I suggest we link to them directly from the main page - maybe a blurb between the first and second paragraphs? Currently we only link to the API and plugins docs.

docs/image-example.rst
jbeezley: Not really.  If anything, it tends to make things more verbose.  I thought about using them, but I didn't really see a benefit in this case.  I think @kotfic is taking another look at the implementation in light of the workflow class.

docs/image-example.rst
jbeezley: No

docs/image-example.rst
kotfic: @jbeezley @jeffbaumes You may be interested in looking at the following task https://github.com/Kitware/minerva/blob/mean-contour-workflow/analyses/NEX/mean_contour_workflow/classes.py#L157-L194  which is apart of a workflow that looks like this:

```python
    wf = Workflow()

    wf.add_task(NetCDFMean(), "m")
    wf.add_task(NetCDFToContourJson(), "c")

    wf.connect_tasks("m", "c", {"output": "data"})

    c = GirderClient('localhost', 8081)
    c.authenticate("kotfic", "letmein")

    output = romanesco.run(
        wf,
        inputs={
             "file_path": {
                 'mode': 'girder',
                 'host': 'localhost',
                 'scheme': 'http',
                 'port': 8081,
                 'api_root': '/api/v1',
                 'resource_type': 'file',
                 'id': "55df4fbaf4b1496693a3c464",
                 'type': 'string',
                 'format': 'text',
                 "name": "test.nc",
                 'token': c.token
             },
            "m.variable": {
                "format": "text",
                "data": "pr"},
            "c.variable": {
                "format": "text",
                "data": "pr"}},
        outputs={
            'contour': {
                'mode': 'girder',
                'name': 'test.json',
                'host': 'localhost',
                'scheme': 'http',
                'port': 8081,
                'api_root': '/api/v1',
                'parent_type': 'folder',
                'parent_id': "55b13466f4b149110f800aa1",
                'format': 'json',
                'type': 'string',
                'token': c.token
            }})

```

This relies on classes defined here: https://github.com/Kitware/minerva/blob/mean-contour-workflow/analyses/NEX/base/__init__.py  which encapsulate the task script handling.  This allows classes inheriting from ```MinervaTask``` to define a ```run``` function with arity equal to the number of input ports. This is still very much a work in progress.

jeffbaumes:
@jbeezley ready for final review? @kotfic: cool, sounds like a good mechanism for specifying analysis code in Python. Are you using a code serialization library? @ronichoudhury: The run function @kotfic is talking about sounds similar to the tangelo entry point - could we somehow normalize the two so a tangelo endpoint could be run through romanesco remotely automatically, perhaps with a decorator? Not an immediate need but that would be super cool.

jbeezley:
@jeffbaumes Yes, I think everything brought up has been addressed.

jeffbaumes:
@jbeezley Could you put a 1-sentence link to the examples page on the main docs page? Once that is there LGTM.

jeffbaumes:
:clap:

PR #171, opened 2015-10-01
Add tests to facebook example

danlamanna:


docs/facebook-example.rst
```diff
@@ -97,36 +146,78 @@ To make this happen, since we've written the tasks already, we just need to form
...
+   --- beginning: most_popular ---
+   --- finished: most_popular ---
+   --- beginning: find_neighborhood ---
+   --- finished: find_neighborhood ---
+
+.. doctest::
+   :hide:
+
+   >>> open('data.json').read() == open('docs/static/data.json').read()
+   True
```
jbeezley: This test is a little fragile in that it relies on the order of the keys in the serialized object, which is technically undefined.

jeffbaumes: Agreed - can we do a JSON parse and deep comparison?

danlamanna:
Ready for review, @jbeezley @jeffbaumes

jbeezley:
LGTM

PR #175, opened 2015-10-14
Add extensions to validator definitions

danlamanna:
This doesn't touch every validator, but the ones currently supported by Tangelohub.

@jeffbaumes


romanesco/plugins/vtk/converters/geometry/validate_vtkpolydata_serialized.json
```diff
@@ -2,5 +2,6 @@
     "inputs": [{"name": "input", "type": "geometry", "format": "vtkpolydata.serialized"}],
     "outputs": [{"name": "output", "type": "boolean", "format": "boolean"}],
     "script": "output = isinstance(input, (str, unicode))",
+    "extensions": ["vtkpolydata.serialized", "vtk"],
```
jeffbaumes: I think these should just be "vtkpolydata", "vtktable", "vtktree" for the first extension for VTK types. I agree that "vtk" should also be included. I think we should make all extensions not have dots in them.

danlamanna:
@jeffbaumes Take a look.

jeffbaumes:
LGTM

PR #176, opened 2015-10-28
Updating pymongo to stay in sync with girder

jbeezley:


zachmullen:
Good thinking @jbeezley . LGTM once travis is green.

PR #180, opened 2015-11-03
JPEG format

zachmullen:


curtislisle:
Zach,
You may have fixed this already, but Jeff and I noticed there was missing unicode options in the validators for PNG and PNG-base64.  I was able to create analyses that write out PNG but couldn’t read the images into a next step if using the Flow interface.  It worked in automated workflows only.  Unicode was the problem.  I was not able to fix and create a pull request myself (permissions problems even after we added me to the repo),  so while you are close to the image validators, could you kindly look at the PNG ones to make sure they accept both string and unicode?  Thanks. 

Curt

jeffbaumes:
There is a way to create a PR from your own fork of romanesco, but I've also added you as a collaborator on romanesco and flow so you can push PRs directly now.

curtislisle:
I have created a “fixPngValidator” branch and will try to push it.

curtislisle:
thanks, Guys, for hand-holding an old colleague in the ways of ‘git’. :-)  PR created.  Please see my note about the extensions attribute on png-base64.

jeffbaumes:
This branch LGTM.

PR #182, opened 2015-11-03
Fix png validator

curtislisle:
Fixed validator on PNG and PNG-base64 to include unicode option so romanesco analyses generating png images can be executed through the Flow editor.  

NOTE:  The PNG-base64 validator may still need an "extensions attribute" to match the PNG one.  I didn't do this because I am uncertain about how the extensions attribute is being used.


jeffbaumes:
LGTM. Go ahead and merge when ready @curtislisle

PR #184, opened 2015-11-04
Scala

jeffbaumes:
Add simple scala mode.

The test uses an inline file which I want to send in as a temporary file, so I also updated the inline io mode to support filepath targeting.


jeffbaumes:
@zachmullen PTAL when you have some time. In particular I want a sanity check on my changes to `romanesco/__init__.py` (always use the io fetch function even in trivial inline case) and `romanesco/io/__init__.py` (add filepath target support).

zachmullen:
The changes to those two files look good to me, but we should add a test for `mode=inline, target=filepath` that asserts the content was written to disk, in both the temp file case and the case with the explicitly-passed filename.

romanesco/__init__.py
```diff
@@ -132,8 +131,7 @@ def convert(type, input, output, **kwargs):
         returns the output binding unchanged.
     """
 
-    if "data" not in input:
-        input["data"] = romanesco.io.fetch(input, **kwargs)
+    input["data"] = romanesco.io.fetch(input, **kwargs)
```
zachmullen: Would it be the case that removing this short-circuit causes the same file to be downloaded twice, for example? Like when using `convert` and `validate`, for example?

zachmullen: I have a feeling that may be the case. If so, we should add another key into the input dict like `fetched` or something to indicate that it has already been fetched.

jeffbaumes:
Ok I've added the testing and ensured fetches are not performed multiple times. PTAL @zachmullen

romanesco/__init__.py
```diff
@@ -100,15 +100,13 @@ def isvalid(type, binding, **kwargs):
     :returns: ``True`` if the binding matches the type and format,
         ``False`` otherwise.
     """
-    if "data" not in binding:
-        binding["data"] = romanesco.io.fetch(binding, **kwargs)
```
jeffbaumes: Note: `run` performs a fetch so I do not think it needs to be done here.

romanesco/__init__.py
```diff
@@ -240,14 +240,14 @@ def extractId(spec):
 
             # Convert data
             if auto_convert:
-                converted = romanesco.convert(task_input["type"], d,
-                                              {"format": task_input["format"]}, **kwargs)
+                converted = romanesco.convert(
+                    task_input["type"], d,
+                    {"format": task_input["format"]},
+                    fetch=False, **kwargs)
+
```
jeffbaumes: This is where the extra fetch was happening so I added a `fetch` parameter for `convert`.

zachmullen:
@jeffbaumes could you merge latest master into this?

jeffbaumes:
I'll try :)

PR #186, opened 2015-11-09
Allow girder IO mode to use memory targets

zachmullen:
Before, we only supported filepath targets. For backward compatibility,
they remain the default, but one can now specify target="memory" for
fetching and pushing.

### DIFF
--- a/romanesco/__init__.py
+++ b/romanesco/__init__.py
@@ -288,7 +288,7 @@ def run(task, inputs=None, outputs=None, auto_convert=True, validate=True,
             if auto_convert:
                 outputs[name] = romanesco.convert(
                     task_output["type"], script_output, d,
-                    task_output=task_output, **kwargs)
+                    **dict({'task_output': task_output}, **kwargs))
             elif d["format"] == task_output["format"]:
                 data = d["script_data"]
                 if d.get("mode"):
--- a/romanesco/__init__.py
+++ b/romanesco/__init__.py
@@ -147,7 +147,7 @@ def convert(type, input, output, **kwargs):
         data = data_descriptor["data"]
 
     if "mode" in output:
-        romanesco.io.push(data, output)
+        romanesco.io.push(data, output, **kwargs)
     else:
         output["data"] = data
     return output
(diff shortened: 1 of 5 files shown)


zachmullen:
So I had a mental error and totally forgot about #161, which was actually an attempt to do this same thing. However, this version of the change supports both filepath and memory targets, so I think we should go ahead and use this one and close the other PR.

romanesco/__init__.py
```diff
@@ -240,14 +241,15 @@ def extractId(spec):
...
+                converted = romanesco.convert(
+                    task_input["type"], d, {"format": task_input["format"]},
+                    **dict({'task_input': task_input}, **kwargs))
                 d["script_data"] = converted["data"]
             elif (d.get("format", task_input.get("format")) ==
                   task_input.get("format")):
                 if "data" not in d:
                     d["data"] = romanesco.io.fetch(
-                        d, task_input=task_input, **kwargs)
+                        d, **dict({'task_input': task_input}, **kwargs))
```
jeffbaumes: I'm unclear how this changes functionality - seems a bit cryptic.

romanesco/__init__.py
```diff
@@ -284,11 +287,13 @@ def extractId(spec):
 
             if auto_convert:
                 outputs[name] = romanesco.convert(
-                    task_output["type"], script_output, d, **kwargs)
+                    task_output["type"], script_output, d,
+                    task_output=task_output, **kwargs)
```
jeffbaumes: This one does not seem consistent with all the other calls to `fetch`, `push`, `isvalid`, `convert` in this file.

kotfic:
@zachmullen if this supersedes https://github.com/Kitware/romanesco/pull/161 feel free to close it.  Fixing that PR got put on the backburner and I never got back around to them.

zachmullen:
@jeffbaumes fixed that issue, and I tried to explain the odd syntax in that last commit message.

romanesco/plugins/girder_io/__init__.py
```diff
@@ -44,27 +48,37 @@ def fetch_handler(spec, **kwargs):
...
+    else:
+        raise Exception('Invalid Girder push target: ' + target)
 
 
 def push_handler(data, spec, **kwargs):
     parent_type = spec.get('parent_type', 'folder')
-    name = spec.get('name', os.path.basename(data))
-    description = spec.get('description')
+    taskOutput = kwargs.get('task_output', {})
+    target = taskOutput.get('target', 'filepath')
```
jeffbaumes: You are setting the default here to filepath - for other modes it seems to make sense to have memory be the default (see my PR at https://github.com/Kitware/romanesco/pull/184/files#diff-792252c08973496d9a93bda90b861811R14). Are we ok with explaining this difference in defaults in the docs?

zachmullen: I'll add something to the docs. I'd love for the default to be consistent, but changing it here would break existing code that uses the girder IO mode.

jeffbaumes:
After docs clarification LGTM.

PR #187, opened 2015-11-10
Detail docs re installing from source

danlamanna:
Leaving this here for @mgrauer to review

### DIFF
--- a/docs/developer-docs.rst
+++ b/docs/developer-docs.rst
@@ -15,7 +15,7 @@ Install requirements: ::
 
     pip install -r requirements.txt
     
-    # If you plan on developing for Romanesco, you may also want to 
+    # If you plan on developing for Romanesco, you may also want to
     pip install -r requirements-dev.txt
 
 Test it: ::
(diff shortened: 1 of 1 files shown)


mgrauer:
Is the entire list of plugins (those that need to be installed like `pip install -e . [spark]`) listed in the plugins.rst file?  I think so, but if not, this should be made explicit.

It might be good to include a link to the plugin docs page from the installation docs page and the developer docs page.  As someone less familiar with Romanesco, I wasn't sure of what I needed to do, the following info from the plugins docs page tells me exactly what I was looking for and couldn't find a few months back.  I think it should have more prominent display, or repetition, or linking, in the docs.


>Adding new execution modes. Without any plugins enabled, the core Romanesco application can only perform two types of tasks: python and workflow modes. It's common for plugins to implement other task execution modes.

danlamanna:
> Is the entire list of plugins (those that need to be installed like pip install -e . [spark]) listed in the plugins.rst file? I think so, but if not, this should be made explicit.

It's not, opening an issue for documentation on swift plugin.

mgrauer:
LGTM.

PR #189, opened 2015-11-10
Girder IO utils

zachmullen:
This depends on #186 .

This adds some much-needed utilities to the server-side plugin to reduce boilerplate for downstream girder plugins that need to build tasks with girder IO dependencies. Also, ported the girder IO mode to support the new `apiUrl` method of instantiating a `GirderClient`.

### DIFF
--- a/docs/plugins.rst
+++ b/docs/plugins.rst
@@ -108,6 +108,7 @@ new item with the same name as the file), or into an existing item.
 
     <GIRDER_OUTPUT> ::= {
         "mode": "girder",
+        "token": <girder token used for authentication>,
         "parent_id": <the _id value of the folder or item to upload into>,
         "format": "text",
         "type": "string"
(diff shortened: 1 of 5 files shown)


zachmullen:
@Kitware/arborworkflows (using this group since it represents those with write access to the repo) could a Romanesco developer please review these two branches? A girder plugin @dzenanz and I are writing for PET/CT calibration depends on these changes. :smile:

zachmullen:
I rebased this on the latest master so this topic is now isolated from its predecessor. PTAL.

jeffbaumes:
Can you add testing to ensure behavior matches docs in terms of defaults, dependencies between keys? Should be easy to test.

zachmullen:
The [existing test](https://github.com/Kitware/romanesco/blob/master/romanesco/plugins/girder_io/tests/girder_io_test.py) (which was improved on #186) already verifies that an unspecified target value defaults to the filepath behavior, and adds tests to ensure that `target="memory"` works as expected. Is there something beyond that that we should test?

zachmullen:
Sorry, I think I misunderstood... were you talking about the keys for specifying the girder URL?

zachmullen:
@jeffbaumes added testing, PTAL

PR #192, opened 2015-11-19
Make scala mode work like Python and R with variable injection

jeffbaumes:
See the test for an example of how this works.

The code now surrounds the Scala script with setup/teardown to allow the use of input variables in the script, and writes outputs to temporary files using `toString()` in Scala. This will enable string, number, and boolean types to be passed into and out of the Scala mode. I wanted to perform generic JSON serialization to handle things like Scala Map and List, but there is no built-in JSON serialization library for Scala.

### DIFF
--- a/romanesco/plugins/scala/executor.py
+++ b/romanesco/plugins/scala/executor.py
@@ -1,38 +1,36 @@
-import re
 import romanesco.utils
-import tempfile
+import os
+import json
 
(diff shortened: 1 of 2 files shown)


jeffbaumes:
@kotfic would you mind taking a look to see if this seems reasonable and potentially useful?

PR #193, opened 2015-11-19
Spark scala support

jeffbaumes:
Add `"spark.scala"` mode which is almost exactly like the `"scala"` mode but uses the `spark-shell` command instead of the `scala` command to execute the code. Also required an `exit(0)` call because `spark-shell` does not normally terminate on script completion.

Depends on #192.

### DIFF
--- a/romanesco/plugins/scala/executor.py
+++ b/romanesco/plugins/scala/executor.py
@@ -15,7 +15,7 @@ def _write_scala_script(script, inputs, task_outputs, tmp_dir):
         script_file.write(script)
 
         # Write output values to temporary files
-        script_file.write('import java.io._')
+        script_file.write('import java.io._\n')
         for name in task_outputs:
             if name != '_stderr' and name != '_stdout':
                 fname = os.path.join(tmp_dir, name)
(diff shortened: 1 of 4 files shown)


jeffbaumes:
@kotfic could you look at this one too?

aashish24:
@jeffbaumes in case @kotfic unable to get to it, I am happy to review as well :smile: later today.

jeffbaumes:
Thanks @aashish24. One larger issue I ignore is packaging dependencies for Spark scala jobs - hopefully we can do something with this in the future.

aashish24:
> packaging dependencies for Spark scala jobs - hopefully we can do something with this in the future.

@jeffbaumes my understanding is that we have this as a general problem in Romanesco, isn't it?

jeffbaumes:
True @aashish24

romanesco/plugins/scala/executor.py
```diff
@@ -1,38 +1,40 @@
...
+        script_file.write(script)
 
-        newArgs.append(arg)
+        # Write output values to temporary files
+        script_file.write('import java.io._\n')
+        for name in task_outputs:
+            if name != '_stderr' and name != '_stdout':
+                fname = os.path.join(tmp_dir, name)
+                script_file.write("""
+new PrintWriter({}) {{
```
aashish24: Curious about empty curly braces!

aashish24:
Looks reasonable to me, just has a minor question on write for outputs

romanesco/plugins/scala/executor.py
jeffbaumes: This is simply a placeholder for the python `format` function - it will be replaced by the name of the temporary file to be written to.

jeffbaumes:
I'm going to translate @aashish24's "reasonable" as "good", which let's me merge this branch.

romanesco/plugins/scala/executor.py
aashish24: > This is simply a placeholder for the python format function - it will be replaced by the name of the temporary file to be written to.

Ah.. makes sense.. thanks.

aashish24:
:+1:

PR #196, opened 2015-11-23
Remove unnecessary assertions that cause adding new types to fail

danlamanna:
@zachmullen


zachmullen:
LGTM

jeffbaumes:
What, we don't like global lists? :) +1 LGTM

PR #198, opened 2015-12-13
Julia execution mode

jeffbaumes:


jeffbaumes:
What's missing is an install of Julia in the Travis environment. There seems to be no simple solution, because Travis can only accept one language. If I could do `language: julia` it would just be there. I also do not see any trivial ways of getting Julia installed as non-root other than building Julia, which we do not want to do. Perhaps we could make an Ubuntu build, store it in a tarball, and wget it onto Travis?

jbeezley:
I would try the generic linux binary download here: http://julialang.org/downloads/.

jeffbaumes:
Thanks for the tip @jbeezley - Julia is now installed. Anyone want to take a look, try it out, and sanity check the test Julia code? This works much like the Swift and Scala plugins - it writes a script wrapping the task script with setup and teardown logic, runs the interpreter which will write temporary files to disk for the outputs, then reads those files into the Romanesco output object.

curtislisle:
Cool!

romanesco/plugins/julia/executor.py
```diff
@@ -0,0 +1,63 @@
+import romanesco.utils
+import os
+import json
+
+
+def _write_julia_script(script, inputs, task_outputs, tmp_dir):
+    script_fname = os.path.join(tmp_dir, 'script.julia')
```
danlamanna: Is there a reason these never get deleted outside of the testing environment?

.travis.yml
```diff
@@ -61,6 +61,10 @@ before_install:
   - export PATH=$PATH:$HOME/swift-0.96.1/bin
   - swift --version || curl -L http://swiftlang.org/packages/swift-0.96.1.tar.gz | tar zx -C ~
   - swift --version
+  # install julia
+  - export JULIA_INSTALL_HOME=$HOME/julia-0.4.2
+  - mkdir $JULIA_INSTALL_HOME && curl https://julialang.s3.amazonaws.com/bin/linux/x64/0.4/julia-0.4.2-linux-x86_64.tar.gz | tar zx -C $JULIA_INSTALL_HOME --strip-components 1
+  - export PATH=$PATH:$JULIA_INSTALL_HOME/bin
```
danlamanna: Should we add this to the travis cache directories?

danlamanna:
This LGTM/works aside from my 2 nitpicks.

jeffbaumes:
@danlamanna I think I've addressed your issues. I was seeing the input text file not being cleaned up, not the script file. See if it works now for you. For me all that is left is an empty `tmp` directory after running the following Julia example script outside of testing:

```
import romanesco

task = {
    'mode': 'julia',
    'script': """
message = "Hello, $foo."
square = x * x
not_b = !b
f = open(file)
lines = readlines(f)
counter = 1
for l in lines
   print("$counter. $l")
   counter += 1
end
close(f)
""",
    'inputs': [
        {
            'id': 'foo',
            'format': 'text',
            'type': 'string'
        },
        {
            'id': 'x',
            'format': 'number',
            'type': 'number'
        },
        {
            'id': 'file',
            'type': 'string',
            'format': 'text',
            'target': 'filepath'
        },
        {
            'id': 'b',
            'format': 'boolean',
            'type': 'boolean'
        }
    ],
    'outputs': [
        {
            'id': '_stdout',
            'format': 'text',
            'type': 'string'
        },
        {
            'id': 'message',
            'format': 'text',
            'type': 'string'
        },
        {
            'id': 'square',
            'format': 'number',
            'type': 'number'
        },
        {
            'id': 'not_b',
            'format': 'boolean',
            'type': 'boolean'
        }
    ]
}

inputs = {
    'foo': {
        'format': 'text',
        'data': 'world'
    },
    'file': {
        'format': 'text',
        'data': """It was the best of times
It was the worst of times
"""
    },
    'x': {
        'format': 'number',
        'data': 12
    },
    'b': {
        'format': 'boolean',
        'data': True
    }
}

out = romanesco.run(task, inputs=inputs)

print out
print {
    '_stdout': {
        'data': """1. It was the best of times
2. It was the worst of times
""",
        'format': 'text'
    },
    'message': {
        'data': 'Hello, world.',
        'format': 'text'
    },
    'square': {
        'data': 144,
        'format': 'number'
    },
    'not_b': {
        'data': False,
        'format': 'boolean'
    }
}
```

danlamanna:
Works! LGTM once travis is happy.

jeffbaumes:
Travis is now happy. Merging.

PR #200, opened 2015-12-31
Add several new job statuses

zachmullen:
We now send separate status flags for fetching, converting, and
pushing. This will help for tracking job performance in many
deployments.

### DIFF
--- a/romanesco/__main__.py
+++ b/romanesco/__main__.py
@@ -1,7 +1,7 @@
 import romanesco
 from romanesco.format import conv_graph
 
-from .utils import JobManager
+from .utils import JobManager, JobStatus
 from celery import Celery
 
 app = None
@@ -23,6 +23,7 @@ def main():
                         url=jobInfo.get('url'), method=jobInfo.get('method'),
                         headers=jobInfo.get('headers')) as jm:
             kwargs['_job_manager'] = jm
+            kwargs['status'] = JobStatus.RUNNING
             retval = romanesco.run(*pargs, **kwargs)
         return retval
 
--- a/tests/io_test.py
+++ b/tests/io_test.py
@@ -276,7 +276,7 @@ with open(file) as f:
                 raise Exception('Unexpected url ' + repr(url))
 
         with httmock.HTTMock(fetchMock):
-            out = romanesco.run(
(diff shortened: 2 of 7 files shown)


romanesco/__main__.py
```diff
@@ -23,6 +23,7 @@ def run(*pargs, **kwargs):
                         url=jobInfo.get('url'), method=jobInfo.get('method'),
                         headers=jobInfo.get('headers')) as jm:
             kwargs['_job_manager'] = jm
```
jeffbaumes: I realize this is older code, but why do we prefix the job manager arg with an underscore? Does it need to be that hidden? Plus it does not match the corresponding status arg that you just now added.

zachmullen: We do that since it is passed as a magic variable into the script env in python mode. It doesn't necessarily need to be underscore prefixed here, though.

jeffbaumes: Ok. Not sure it should be prefixed here or in the the script env - e.g. we inject the Spark context as `sc` with no underscore in the spark plugin. I'll create a separate issue to address this - I wouldn't change anything in this PR.

jeffbaumes:
Could we add a simple test that does input/output conversion and verifies the job status is updated?

zachmullen:
Yep, I will add that.

zachmullen:
This is now ready for review.

tests/io_test.py
```diff
@@ -195,6 +214,7 @@ def fetchMock(url, request):
                 task, inputs=copy.deepcopy(inputs), cleanup=False,
                 validate=False, auto_convert=False)
             self.assertEqual(out['y']['data'], 'dummy file contents_suffix')
+            """
```
jeffbaumes: Shall we remove this testing that is now commented out or were you going to fix it up and keep it?

jeffbaumes:
After Travis is happy (I think just some style issues?) and one small note on commented-out code, I think this will be good to go.

zachmullen:
Whoops, thanks for catching that, I left out a commit on my rebase. Should be fixed now.

jeffbaumes:
LGTMITIH

PR #203, opened 2016-01-16
Upgrade Girder, Minerva, Romanesco common requirements

### LINKED ISSUE #194: bump requests version
Would Romanesco be able to change its requests version to

    requests==2.7.0

or

    requests>=2.6.0

?

I'm running into trouble with incompatible requests versions between girder[requests==2.7.0], romanesco[requests==2.6.0] and minerva[girder-client => requests==2.7.0], while trying to document an install path using as much pip/girder-install as possible.

Current workflow is 

    sudo pip install girder
    git clone romanesco to romanesco_dir
    git clone minerva to minerva_dir
    create worker.local.cfg with spark plugin enabled in romanesco_dir/romanesco
    sudo pip -e .[spark] in romanesco_dir
    sudo girder-install plugin romanesco_dir
    sudo girder-install plugin minerva_dir

brianhelba:
Fixes #194.

amqp changes at (not up to date):
https://github.com/celery/py-amqp/blob/master/Changelog

Billiard changes at (not up to date):
https://github.com/celery/billiard/blob/master/CHANGES.txt

Kombu changes at:
http://docs.celeryproject.org/projects/kombu/en/latest/changelog.html

Pillow changes at:
https://pillow.readthedocs.org/en/3.1.x/releasenotes/3.1.0.html

PyMongo changes at:
https://api.mongodb.org/python/current/changelog.html

pytz changes at (not up to date):
https://launchpad.net/pytz/+announcements

Requests changes at:
http://docs.python-requests.org/en/latest/community/updates/#release-and-version-history

Six changes at:
https://bitbucket.org/gutworth/six/src/1991f8b5b654f077e773f05695a08e0506b7367f/CHANGES


brianhelba:
This should be merged along with: https://github.com/Kitware/minerva/pull/259 and https://github.com/girder/girder/pull/11773 to bring all three projects in sync and up to date (with respect to common requirements only).

brianhelba:
Fixed a test failure caused by Pillow upgrade. In the test, the new version of Pillow generates an output image with a slightly different mean pixel value.

I'm not sure why this happens, to be honest.

brianhelba:
Here's why the image test needed to be updated:

From the Pillow 2.7.0 release notes:
> The GaussianBlur() implementation has been replaced with a sequential
application of box filters.

This results in a slightly different pixel mean value after blurring.

> There was an error in the previous version of Pillow, where blur radius
(the standard deviation of Gaussian) actually meant blur diameter. For
example, to blur an image with actual radius 5 you were forced to use
value 10. This has been fixed. Now the meaning of the radius is the same
as in other software.

> If you used a Gaussian blur with some radius value, you need to divide
this value by two.

This requires the input "blur_radius" parameters to be updated in the test.

brianhelba:
I also fixed a test failure that appears to also be causing problems on the existing master. I think it's cropped up in the last 11 days, since master was last updated. If someone could confirm this (I don't have Romanesco set up to test locally, and I don't know how / have permission to force Travis to re-run tests on the current master), I'd appreciate it.

One of the tests makes calls to api.opentreeoflife.org. This service seems to now require the use of HTTPS.

Despite the fact that the Requests library is used to make these calls, this issue appears to be unrelated to the upgrade of this library (the failure occurs even with the prior version of the library).

mgrauer:
This required system dependencies `libjpeg-dev` `zlib1g-dev` on Ubuntu 14.04.  This should be documented.

curtislisle:
The API for Opentree has been evolving. I am not surprised if our test is out of date. As far as. Am concerned, we could drop a test that directly invokes open tree in favor of test for with a small tree file instead. Romensco tests should break it Opentree changes their API.

brianhelba:
Ping @jeffbaumes . I believe he wrote the tests in question.

jeffbaumes:
Agreed that the opentree test should go away.

brianhelba:
@jeffbaumes @mgrauer @zachmullen Ping.

mgrauer:
@brianhelba  Can you also add those system packages to the installation.rst or link back from there?

I realize there may be a need for larger doc grooming, but I think this is a reasonable addition within this PR.

brianhelba:
@mgrauer Updated my last commit to add the note in ``installation.rst``.

mgrauer:
LG from my end

brianhelba:
Merging.

PR #206, opened 2016-02-02
Call Resource constructor from Validator, fixing a Girder warning

brianhelba:


brianhelba:
@zachmullen PTAL.

zachmullen:
LGTM

